PR #2, opened 2021-11-30
Open `CHE_PROJECTS_ROOT` by default

### LINKED ISSUE #20569: Add `/projects` subfolders to VS Code explorer when opening the workspace

Currently, when I start my workspace with Che-Code as an editor, there's no folder opened. I need to open `/projects` manually to open my imported project sources.

image

Need to investigate how to configure Che-Code editor to open `/projects` subfolders as the workspace folders.

_No response_

### Additional context

This issue is a part of the epic #20500

azatsarynnyy:
Signed-off-by: Artem Zatsarynnyi <azatsary@redhat.com>

closes https://github.com/eclipse/che/issues/20569

### DIFF
--- a/build/scripts/entrypoint-volume.sh
+++ b/build/scripts/entrypoint-volume.sh
@@ -29,4 +29,4 @@ else
 fi
 
 # Launch che with a custom connection-token
-./node out/vs/server/main.js --port 3100 --connection-token eclipse-che
+./node out/vs/server/main.js --port 3100 --connection-token eclipse-che --folder ${CHE_PROJECTS_ROOT}
--- a/build/scripts/entrypoint-volume.sh
+++ b/build/scripts/entrypoint-volume.sh
@@ -29,4 +29,4 @@ else
 fi
 
 # Launch che with a custom connection-token
-./node out/vs/server/main.js --port 3100 --connection-token eclipse-che --folder ${CHE_PROJECTS_ROOT}
+./node out/vs/server/main.js --port 3100 --connection-token eclipse-che --folder ${PROJECTS_ROOT}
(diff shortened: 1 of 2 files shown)


build/scripts/entrypoint-volume.sh
```diff
@@ -29,4 +29,4 @@ else
 fi
 
 # Launch che with a custom connection-token
-./node out/vs/server/main.js --port 3100 --connection-token eclipse-che
+./node out/vs/server/main.js --port 3100 --connection-token eclipse-che --folder ${CHE_PROJECTS_ROOT}
```
benoitf: AFAIK CHE_PROJECTS_ROOT is only available for che-server workspaces

azatsarynnyy: You're right. It's `PROJECTS_ROOT` for the DevWorkspace engine. I forgot about that.
I'm updating the PR...
Thanks for noticing it!

benoitf: Also I think it should open the folders defined by the projects of the devfile.yaml
Here it might open the root folder which may not what we expect.
I'm fine to merge this one as it's still better than before but probably we need to do the same stuff than for che-theia where we create a .workspace file and open that .workspace file

Now it's quite easy as projects are cloned before we launch the IDE (in devWorkspace mode)

azatsarynnyy: As I remember, Che-Code editor is supposed to work with DevWorkspace engine only. So, I've replaced `CHE_PROJECTS_ROOT` with `PROJECTS_ROOT`.

azatsarynnyy: > Also I think it should open the folders defined by the projects of the devfile.yaml

We can add the custom folders support as one of further improvements.

benoitf (approved):

benoitf:
Further improvement https://github.com/eclipse/che/issues/20870

PR #22, opened 2022-01-28
fix: change default behaviour on unimplemented methods

nils-mosbach:
Checked with my environment. This change as recommended by @benoitf solves the issue.

https://github.com/eclipse/che/issues/21081


code/src/vs/server/node/che/remoteTerminalMachineExecChannel.ts
```diff
@@ -158,7 +158,7 @@ export class RemoteTerminalMachineExecChannel implements IServerChannel<RemoteAg
 		}
 
 		this.logService.error(`RemoteTerminalChannel: unsupported command/${command}`);
-		return undefined;
+		return {};
```
benoitf, line 161: I think we should return {} only if command name is $getEnvironment to address the real issue that was for getEnvironment but why not return an empty object instead undefined for unsupported case anyway

benoitf (approved):

benoitf:
thank you for the fix, @nils-mosbach

PR #23, opened 2022-01-31
feat: enhance user experience

benoitf:
- open terminal by default (if no terminal is there)
- add the remote customization: text, icon, commands available when clicking on the bottom left remote

https://user-images.githubusercontent.com/436777/151932251-d30da57a-27f2-4198-b697-92e109bf5d75.mp4


code/extensions/che-on-start/README.md
```diff
@@ -0,0 +1,7 @@
+# Eclipse Che Port for Visual Studio Code
```
azatsarynnyy: Please, update the readme

code/extensions/che-on-start/package.nls.json
```diff
@@ -0,0 +1,4 @@
+{
+	"displayName": "Eclipse Che OnStart",
+	"description": "OnStart actions like project automatically at startup if no project is provided"
```
azatsarynnyy: Could you please rephrase it a bit, as it's hard to understand the extension's description

code/extensions/che-remote/__mocks__/vscode.ts
```diff
@@ -0,0 +1,45 @@
+/* eslint-disable header/header */
```
azatsarynnyy: Why do we need that rule here?

code/extensions/che-on-start/src/extension.ts
```diff
@@ -13,7 +13,24 @@
...
+  const output = vscode.window.createOutputChannel('Che OnStart');
+  openTerminalIfNone(output);
+  openProjectIfNone(output);
+
+}
+
+async function openTerminalIfNone(output: vscode.OutputChannel): Promise<void> {
+  // open a new terminal if there is none to always see a terminal
+  if (vscode.window.terminals.length === 0) {
+    output.appendLine('Opening a new terminal....');
```
azatsarynnyy, line 32: `output` is declared in another function - `activate`.
What `output` does it use here?

benoitf: it's using the output given in parameter initialized in activate method ?

code/extensions/che-remote/__mocks__/vscode.ts
benoitf: we have the rule in every che files else it tries to validate with Microsoft license

azatsarynnyy: I see. Thanks for the explanation!

benoitf: but I'll move this line after the header
At first I thought it was only working at the beginning of the file

And I think that probably we could tweak linter rule to ignore the che-* extensions

code/extensions/che-on-start/src/extension.ts
azatsarynnyy: ahh... I overlooked it's passed through the parameter. Sorry for the noise )

code/extensions/che-on-start/README.md
benoitf: updated

code/extensions/che-on-start/package.nls.json
benoitf: updated

code/extensions/che-remote/__mocks__/vscode.ts
benoitf: moved the line to the bottom

benoitf:
ok, so opening a default terminal by default is great but needs to be disabled in unit test/smoke test mode else we had random failures ! :-)

PR #26, opened 2022-02-01
fix: get rid of the connectionToken parameter

### LINKED ISSUE #21099: che-code - Open IDE redirects too fast: `Bad Gateway`
### Describe the bug

If starting a new Workspace opening the IDE e.g. che-code is triggered to fast and responds with a `Bad Gateway`. If clicked refresh in the browser after 2 seconds everything loads as expected. Maybe che-code or the traeffik configration is not yet ready? 

So basically the redirect from here

image

to the IDE happens to fast.

image

Startup times of workspaces are incredibly fast after we switched to devworkspaces. Great improvement! :)

### Che version

next (development version)

### Steps to reproduce

Start che-code environment.

### Expected behavior

Che code should be opened as soon as its ready.

### Runtime

Kubernetes (vanilla)

### Screenshots

_No response_

### Installation method

chectl/latest

### Environment

Linux

### Eclipse Che Logs

_No response_

### Additional context

_No response_

benoitf:
Che is securing the route so we don't need connectionToken parameter
It's now possible to turn off this feature in VS Code

it might fix https://github.com/eclipse/che/issues/21099 as the gateway can know check if the server is returning 200 and it's no longer returns 403 error

### DIFF
--- a/build/scripts/entrypoint-volume.sh
+++ b/build/scripts/entrypoint-volume.sh
@@ -29,4 +29,4 @@ else
 fi
 
 # Launch che with a custom connection-token
-./node out/server-main.js --port 3100 --connection-token eclipse-che
+./node out/server-main.js --port 3100 --without-connection-token


azatsarynnyy (approved):

benoitf:
so while it's getting rid of the connectionToken I still have bad gateway issue

PR #39, opened 2022-04-07
chore: Apply Che GitHub authentication provider instead of built-in

vinokurig:
- Remove the built-in github-authentication extention.
- Add Che specific github-authentication extension.
Signed-off-by: Igor Vinokur <ivinokur@redhat.com>


benoitf:
hello, is there a way to let the other plug-in in the source code but not enable it ?
like specifying it there:
https://github.com/che-incubator/che-code/blob/main/code/build/lib/extensions.ts#L235-L242 (+ counterpart in extensions.js)

I think it might break the automatic rebase to have it removed while if it's just excluded it's safer/easier to maintain.

vinokurig:
@benoitf
> hello, is there a way to let the other plug-in in the source code but not enable it ?
like specifying it there:
https://github.com/che-incubator/che-code/blob/main/code/build/lib/extensions.ts#L235-L242 (+ counterpart in extensions.js)
>
> I think it might break the automatic rebase to have it removed while if it's just excluded it's safer/easier to maintain.

Excluding the original `github-authentication` extension in the `code/build/lib/extensions` files doesn't prevent the extension from start, moreover, an error appears:
```
Activating extension 'vscode.github-authentication' failed: Cannot find module '/Users/ivinokur/projects/tmp/che-code/code/extensions/github-authentication/out/extension.js' Require stack: - /Users/ivinokur/projects/tmp/che-code/code/out/vs/loader.js - /Users/ivinokur/projects/tmp/che-code/code/out/bootstrap-amd.js - /Users/ivinokur/projects/tmp/che-code/code/out/bootstrap-fork.js.
```
Is there another way to exclude the extension?

benoitf:
@vinokurig you may try there
https://github.com/che-incubator/che-code/blob/5c33cee7ef4ca81d85cac6a0642dd2822cb910ec/code/src/vs/server/node/remoteAgentEnvironmentImpl.ts#L353-L355

https://github.com/che-incubator/che-code/blob/5c33cee7ef4ca81d85cac6a0642dd2822cb910ec/code/src/vs/workbench/services/extensionManagement/browser/builtinExtensionsScannerService.ts

vinokurig:
@benoitf 
Hello, I added some tests for the new extension and a filter of excluded extensions to the `extensionsScannerService.ts`
Could you please take a look?

code/extensions/che-github-authentication/package.json
```diff
@@ -0,0 +1,80 @@
...
+    "vscode-tas-client": "^0.1.31"
+  },
+  "devDependencies": {
+    "@types/node": "16.x",
+    "@types/node-fetch": "^2.5.7",
+    "@types/uuid": "8.0.0"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/microsoft/vscode.git"
```
benoitf: repository looks invalid

benoitf:
@vinokurig I don't see any changes to a file named `extensionsScannerService` in that PR ?

vinokurig:
@benoitf
> I don't see any changes to a file named extensionsScannerService in that PR ?

https://github.com/che-incubator/che-code/pull/39/files#diff-36fae229a2323a58cfac9a248b606846fd67fbf795dd4f8aecacfd6007faf134

code/extensions/che-github-authentication/package.json
vinokurig: fixed

code/src/vs/platform/extensionManagement/common/extensionsScannerService.ts
```diff
@@ -412,6 +412,17 @@ class ExtensionsScanner extends Disposable {
...
 	}
 
+	private readonly excludedExtensions = [
+		'vscode-api-tests',
+		'vscode-colorize-tests',
+		'vscode-test-resolver',
+		'ms-vscode.node-debug',
+		'ms-vscode.node-debug2',
+		'vscode-notebook-tests',
+		'vscode-custom-editor-tests',
```
benoitf: why these one are excluded ?

vinokurig: The list comes from https://github.com/che-incubator/che-code/blob/main/code/build/lib/extensions.ts#L282-L290. I thought that it was added by our team and since it doesn't work for `github-authentication` it doesn't work for others, so I decided to add those extensions as well. Am I correct?

benoitf: it's added by upstream vscode so I would just add there github one

benoitf (approved):

code/src/vs/platform/extensionManagement/common/extensionsScannerService.ts
vinokurig: done

benoitf:
@vinokurig thanks !

you can probably enable auto-merge so once all tests are green it'll be auto-merged

PR #50, opened 2022-05-16
Update entrypoint volume to support arbitrary user

l0rd:
When using che-code with Che, the entrypoint of the container that will host che-code is overridden and `entrypoint-volume.sh` is executed. This is problematic when the container is run as an arbitrary user and files `/etc/passwd` and `/etc/group` are supposed to be patched. See screenshot below.

image

To fix that I have added the boilerplate code that patches `/etc/passwd` and `/etc/group` in the script `entrypoint-volume.sh`.

### DIFF
--- a/build/scripts/entrypoint-volume.sh
+++ b/build/scripts/entrypoint-volume.sh
@@ -11,6 +11,14 @@
 #   Red Hat, Inc. - initial API and implementation
 #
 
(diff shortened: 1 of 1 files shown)


benoitf (approved):
LGTM

l0rd:
I have tested the PR image and it works well and fix the problem.

PR #57, opened 2022-06-24
Optimize running the `pr-check` workflow

azatsarynnyy:
Don't start `pr-check` workflow if the changes don't affect the build.
It's the second part. The first one is https://github.com/che-incubator/che-code/pull/59

closes https://github.com/eclipse/che/issues/21495

PR #59, opened 2022-06-24
Add a special workflow to skip the mandatory pr-check on the PRs with the changes in the build non-related files

azatsarynnyy:
Signed-off-by: Artem Zatsarynnyi <azatsary@redhat.com>

The PR adds the workflow that allows passing the mandatory [`pr-check.yaml`](https://github.com/che-incubator/che-code/blob/main/.github/workflows/pr-check.yml) in case
of the changes made in the files that aren't related to the build.

For more details, see https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks

Part of the fix for https://github.com/eclipse/che/issues/21495

The second part is https://github.com/che-incubator/che-code/pull/57

### DIFF
--- a/.github/workflows/pr-check-ignore.yml
+++ b/.github/workflows/pr-check-ignore.yml
@@ -1,0 +1,31 @@
+#
+# Copyright (c) 2023 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
+# The workflow allows to pass the mandatory pr-check.yaml in case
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-check-ignore.yml
```diff
@@ -0,0 +1,32 @@
+#
+# Copyright (c) 2023 Red Hat, Inc.
```
benoitf: Looks like you're in advance on the date :-)

benoitf (approved):

.github/workflows/pr-check-ignore.yml
azatsarynnyy: I just can't wait for 2023 :smile:

PR #65, opened 2022-07-04
Port @eclipse-che/theia-activity-tracker

dkwon17:
This PR ports [eclipse-che-theia-activity-tracker](https://github.com/eclipse-che/che-theia/tree/main/extensions/eclipse-che-theia-activity-tracker) to che-code.

There were two main differences with https://github.com/eclipse-che/che-theia/tree/main/extensions/eclipse-che-theia-activity-tracker and the VS Code plugin introduced in this PR:

### Tracking
Instead of [front-end tracking like in che-theia](https://github.com/eclipse-che/che-theia/blob/60e8ad8784629294b0c088c039c4472567b11d87/extensions/eclipse-che-theia-activity-tracker/src/browser/che-theia-activity-tracker-contribution.ts#L36-L38), this PR tracks the following VS Code events using the extension API:
* `vscode.workspace.onDidChangeTextDocument`
* `vscode.window.onDidChangeActiveTextEditor`
* `vscode.window.onDidChangeTextEditorSelection`
* `vscode.window.onDidChangeTextEditorViewColumn`
* `vscode.window.onDidChangeWindowState`
* `vscode.window.onDidChangeTerminalState`
* `vscode.window.onDidChangeActiveTerminal`

### Determining the che-machine-exec url
Instead using an `EndpointService` to determine the machine-exec port like for [che-theia](https://github.com/eclipse-che/che-theia/blob/60e8ad8784629294b0c088c039c4472567b11d87/extensions/eclipse-che-theia-remote-impl-k8s/src/node/k8s-workspace-service-impl.ts#L79), this PR relies on an environment variable set by `entrypoint-volume.sh` which executes the machine-exec binary. The environment variable is read [here](https://github.com/che-incubator/che-code/blob/a954fdbd00f5a93e311be10d65ceae19b04cbf03/code/extensions/che-api/src/impl/k8s-workspace-service-impl.ts#L44).This was done because for Che-code, because for che-code, che-machine-exec is not running in its own container.

## To test this PR:
Since there is no idling set up for Che-code at the moment, to test this PR, create a workspace with this che-code image: `quay.io/dkwon17/che-code:idle`. Here is the editor devfile v2: https://gist.githubusercontent.com/dkwon17/1a96b69643260c9af23553f740132157/raw/72fb81012aa03618490b91cf887ff6a5cd12a889/gistfile1.txt.

For example, to start a golang che sample workspace, go to:
```
{CHE_HOST}/#https://github.com/che-samples/golang-example/tree/devfilev2?che-editor=https://gist.githubusercontent.com/dkwon17/1a96b69643260c9af23553f740132157/raw/72fb81012aa03618490b91cf887ff6a5cd12a889/gistfile1.txt
```

When the workspace is running, the `tools` container of the workspace pod should log `/activity/tick` `POST` messages when interacting with the che-code editor. Requests are sent once a minute at most to che-machine-exec:
image

Signed-off-by: David Kwon <dakwon@redhat.com>


dkwon17:
This PR is for https://github.com/eclipse/che/issues/21513

build/scripts/entrypoint-volume.sh
```diff
@@ -23,7 +23,8 @@ fi
 ls -la /checode/
 
 # Start the machine-exec component in background
-nohup /checode/bin/machine-exec --url '0.0.0.0:3333' &
+export MACHINE_EXEC_PORT=3333
```
dkwon17, line 26: This was done so that the VS Code extension could determine the machine exec port

code/extensions/che-activity-tracker/src/activity-tracker-service.ts
```diff
@@ -0,0 +1,81 @@
+/**********************************************************************
+ * Copyright (c) 2019-2022 Red Hat, Inc.
```
azatsarynnyy: Copyright (c) 2022

build/scripts/entrypoint-volume.sh
```diff
@@ -23,7 +23,8 @@ fi
...
+nohup /checode/bin/machine-exec --url "0.0.0.0:${MACHINE_EXEC_PORT}" &
```
azatsarynnyy, line 27: Che-Code will start sending the user the activity.
So, now it makes sense to enable the automatic stop of the inactive workspaces:
`--idle-timeout ${DEVWORKSPACE_IDLE_TIMEOUT}`
https://github.com/eclipse-che/che-machine-exec/blob/302ec2e92350bf77c81d4d66a8be0e0a843d1f93/cfg/cfg.go#L85
WDYT?

code/extensions/che-activity-tracker/src/activity-tracker-service.ts
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ ***********************************************************************/
+
+/**
+ * Receives activity updates and sends reset inactivity requests to the che-exec /activity/tick endpoint.
+ * To avoid duplicate requests may send requests periodically. This means
+ * that, in the worst case, it might keep user's workspace alive for a longer period of time.
+ */
+export class ActivityTrackerServiceImpl {
```
azatsarynnyy: maybe without the `Impl` suffix, as there's no interface

code/extensions/che-activity-tracker/src/activity-tracker-service.ts
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ ***********************************************************************/
+
+/**
+ * Receives activity updates and sends reset inactivity requests to the che-exec /activity/tick endpoint.
```
azatsarynnyy: `che-exec` > `machine-exec`

azatsarynnyy (commented):
works well - the endpoint is requested regularly as expected
image

build/scripts/entrypoint-volume.sh
dkwon17: I think we should leave out `--idle-timeout ${DEVWORKSPACE_IDLE_TIMEOUT}` here in favour of https://github.com/eclipse/che/issues/21390,

where the workflow would be:
If `--idle-timeout` is not provided to che-machine-exec, machine-exec would read `SECONDS_OF_DW_INACTIVITY_BEFORE_IDLING` env variable to determine the timeout

code/extensions/che-activity-tracker/src/activity-tracker-service.ts
dkwon17: Sounds good, updated

dkwon17:
Thank you for the review @azatsarynnyy , I've made the changes

build/scripts/entrypoint-volume.sh
azatsarynnyy: Agree :+1:

azatsarynnyy (approved):
Thank you @dkwon17 for this contribution.

top1436:
@dkwon17  Thank you for your explanation. 

"devEnvironments.secondsOfInactivityBeforeIdling" is right. my bad...

If che-code built-in extension checks "Activity", Another IDE cannot be applied by this option?  ex) IntelliJ

In my exprience, Intellij is also applied this option and it stopped Intellij workspace after the seconds.

In IntelliJ IDE, Is it implemented  a diffrent way?

azatsarynnyy:
Hi @top1436 
JetBrains editors in Che have no support yet for sending the user's activity to the machine-exec component that stops the workspaces.

top1436:
@azatsarynnyy hi!
you mean in jetbrains editor, always inactivity state and workspace always stops after secondsOfInactivityBeforeIdling?

azatsarynnyy:
@top1436 correct, if I haven't missed anything )

top1436:
@azatsarynnyy
thank you! I think it needs to be fixed. Is there any plan to develop jetbrain plugin as che-code does?

azatsarynnyy:
> @azatsarynnyy thank you! I think it needs to be fixed. Is there any plan to develop jetbrain plugin as che-code does?

There're no plans to invest in the current (Projector-based) implementation of JetBrains editors for Che.
But we're going to look at JetBrains Gateway instead (https://github.com/eclipse/che/issues/21412)

PR #69, opened 2022-07-12
Publish the `next` images from the dedicated branch

azatsarynnyy:
Publish the `next` images from the dedicated branch that points to a previous commit. It's because the current state of the `main` branch requires some adaptation to the upstream changes.
This PR allows us to get `che-incubator/che-code/insiders` with a working terminal until we investigate and fix the https://github.com/eclipse/che/issues/21537

### DIFF
--- a/.github/workflows/image-publish-insiders.yml
+++ b/.github/workflows/image-publish-insiders.yml
@@ -10,6 +10,7 @@
 name: image-publish-insiders
 
 on:
+  workflow_dispatch:
   push:
     branches: [ trm-fix ]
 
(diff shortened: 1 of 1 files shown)


RomanNikitenko (approved):

nickboldt:
on a related note, is there a stable branch from which to productize che-code into ds-code? 

Right now I only see `main` and some topic branches at https://github.com/che-incubator/che-code/branches 

for DS 3.2 we will NEED a stable branch, be it "7.52.x" or "devspaces-3.2-rhel-8" or some other convention you prefer.

azatsarynnyy:
> on a related note, is there a stable branch from which to productize che-code into ds-code?

@nickboldt  I've created https://github.com/che-incubator/che-code/tree/devspaces-3.2-rhel-8

PR #86, opened 2022-08-04
fix: Fix 'unsupported platform' error for s390x platform

RomanNikitenko:
We have an error for `s390x` platform on `yarn install` step:

image

It's because of `esbuild` version:

image

The upgrade of `esbuild` version fixes the problem:
Screenshot 2022-07-23 at 13 18 48

The change was tested on `s390x` platform, please see details [here](https://issues.redhat.com/browse/CRW-3171?focusedCommentId=20709740&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-20709740) 

Steps for testing `rebase.sh` script changes please see here: https://github.com/che-incubator/che-code/pull/86#issuecomment-1207371544

### DIFF
--- a/.rebase/override/code/build/package.json
+++ b/.rebase/override/code/build/package.json
@@ -1,0 +1,5 @@
+{
+  "devDependencies": {
+    "esbuild": "^0.14.49"
+  }
+}
--- a/.rebase/override/code/extensions/package.json
+++ b/.rebase/override/code/extensions/package.json
@@ -1,0 +1,5 @@
+{
+  "devDependencies": {
+    "esbuild": "^0.14.49"
+  }
+}
--- a/code/build/package.json
+++ b/code/build/package.json
@@ -48,7 +48,7 @@
     "commander": "^7.0.0",
     "debug": "^4.3.2",
     "electron-osx-sign": "^0.4.16",
-    "esbuild": "^0.14.2",
+    "esbuild": "^0.14.49",
     "extract-zip": "^2.0.1",
     "fs-extra": "^9.1.0",
     "got": "11.8.5",
--- a/code/extensions/package.json
+++ b/code/extensions/package.json
@@ -11,7 +11,7 @@
   },
   "devDependencies": {
     "@parcel/watcher": "2.0.5",
-    "esbuild": "^0.11.12",
+    "esbuild": "^0.14.49",
     "vscode-grammar-updater": "^1.1.0"
   }
 }
--- a/rebase.sh
+++ b/rebase.sh
@@ -107,6 +107,28 @@ apply_code_remote_package_changes() {
   git add code/remote/package.json > /dev/null 2>&1
 }
 
+# Apply changes on $1 package.json file
+# A path to the file should be passed, for example: code/build/package.json 
+apply_package_changes_by_path() {
+  local filePath="$1"
+  
+  if [ -z "$filePath" ]; then
+     echo "Can not apply changes for package.json file - the path was not passed"
+     exit 1;
+  fi
+
+  echo "  ⚙️ reworking $filePath..."
+  
+  # reset the file from what is upstream
+  git checkout --theirs $filePath > /dev/null 2>&1
+  
+  # now apply again the changes
+  override_json_file $filePath
+  
+  # resolve the change
+  git add $filePath > /dev/null 2>&1
+}
+
 # Apply changes on code/remote/yarn.lock file
 apply_code_remote_yarn_lock_changes() {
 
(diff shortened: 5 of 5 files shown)


benoitf:
hi, I think this commit will be lost during the self update of code
as it's not in VS Code https://github.com/microsoft/vscode/blob/main/build/package.json#L51

if you need to keep it, you need to add two files there and in the rebase script https://github.com/che-incubator/che-code/tree/main/.rebase/override/code 
so each time the file is rebased, it will update to the values you specified

RomanNikitenko:
> hi, I think this commit will be lost during the self update of code
as it's not in VS Code https://github.com/microsoft/vscode/blob/main/build/package.json#L51

Hello, Florent!
I thought about it and yesterday we discussed with Artem that I should provide some changes to `.rebase` folder.
but :-)
today I played with `rebase.sh` script, tested different cases => I didn't detect that the versions replaced by ones from upstream.

I mean:
- my changes are still present in `package.json` files after applying `rebase.sh` script
- the versions are not replaced and there is no conflict if upstream has a change for the same `package.json` file but another line
- the versions are not replaced automatically but there is a conflict if upstream has a change exactly for the same line - `esbuild`

To clarify - I don't mind to provide some changes for `.rebase` folder, but it looks like they are not necessary...
Please let me know if I missed something.

benoitf:
it'll become necessary when package.json upstream will be modified or there will be new changes or conflicts (in that case it might reset to the original and then we need to apply our changes)

benoitf:
For example let say in upstream someone say : "let's unify the version we have in extensions and take "esbuild": "^0.14.2", instead of 0.11" then when rebasing we don't want conflicts or we don't want that it would fetch the upstream value

RomanNikitenko:
I think it's use case` #3` in my comment.
OK, I see, thank you, Florent for the clarification.

benoitf (approved):

RomanNikitenko:
I think something wrong with [override_json_file()](https://github.com/che-incubator/che-code/blob/160f706bb542eaf15ecd8f1a24692b7b91a0f61d/rebase.sh#L18-L36) function even without my changes.

According to the logic - it's expected to have both:
- `.rebase/override/$filename` 
- `.rebase/add/$filename`

If only `.rebase/override/$filename`  is provided - the function do nothing...
Isn't it?

Investigating...

RomanNikitenko:
Now it should work correctly.

I tested case when we have only `override` values for a `package.json` file.
I'm going to test other use cases, like:
- only `add` values
- `add` and `override` values at the same time

RomanNikitenko:
I tested my changes using auxiliary branches, [origin/testBr is used ](https://github.com/RomanNikitenko/che-code/blob/cfd587b2be78d085fb106bca7cb54f3dafe712b1/rebase.sh#L272) for `rebase`([the changes are here](https://github.com/che-incubator/che-code/compare/main...RomanNikitenko:che-code:testBr?expand=1)), so the steps for testing are:

### `Add` + `Override` use case
- git clone https://github.com/RomanNikitenko/che-code.git
- cd che-code
- git checkout upgradeEsbuildVersion_Copy
- ./rebase.sh

#### Expected behaviour:
- go to `che-code/code/build/package.json`
- check that the `esbuild` version the same as defined in `che-code/.rebase/override/code/build/package.json`: `"esbuild": "^0.14.49"`
- check that a new dependency came from `che-code/.rebase/add/code/build/package.json`: `"someNewDependency": "^1.1.1"`

### `Override` use case
- git reset --hard b35b094eb20e096341c269f407a92056be9175b6
-  now we don't have `che-code/.rebase/add/code/build/package.json` - testing only `override`
- ./rebase.sh

#### Expected behaviour:
- go to `che-code/code/build/package.json`
- check that the `esbuild` version the same as defined in `che-code/.rebase/override/code/build/package.json`: `"esbuild": "^0.14.49"`

### `Add` use case + new changes from upstream
- git reset --hard b35b094eb20e096341c269f407a92056be9175b6
- git pull origin upgradeEsbuildVersion_Copy --rebase
- rm .rebase/override/code/build/package.json
- commit the changes
-  now we don't have `che-code/.rebase/override/code/build/package.json`
- ./rebase.sh

#### Expected behaviour:
- go to `che-code/code/build/package.json`
- check that the new `esbuild` version came from the upstream: `"esbuild": "^0.14.499999999",` - we don't have `overriding`
- check that a new dependency came from `che-code/.rebase/add/code/build/package.json`: `"someNewDependency": "^1.1.1"`

RomanNikitenko:
@benoitf 
I provided some changes for `rebase.sh` script, please take a look if it looks good to you.
thanks in advance!

benoitf (approved):
👍

PR #103, opened 2022-09-12
feat: Add Che Code release workflow

### LINKED ISSUE #21552: Setup the release job for Che-Code

Currently, the [Che-Code](https://github.com/che-incubator/che-code) repository doesn't have a release job that could be included in the overall Che release process.

Create a GitHub Workflow that would:
- run a script that prepares a release branch;
- build the images;
- publish the built images to the Quay registry.

_No response_

### Additional context

_No response_

mkuznyetsov:
closes https://github.com/eclipse/che/issues/21552

New workflow, that will provide a new versioning & release system similar to other Che Projects (7.x.y)
Version will be specified in package.json 
The "che-code-dev" image will not be tagged during release.

UPD: after some considerations with Nick, we decided to keep release.yml containing release "bootstrap" logic, like creating branches, while image-publishing workflow that exists for "insiders/next build", will also be used for building release images, thus avoiding duplicated image building logic for 2 workflows

Signed-off-by: Mykhailo Kuznietsov <mkuznets@redhat.com>


make-release.sh
```diff
@@ -0,0 +1,138 @@
+#!/bin/bash
+#
+# Copyright (c) 2020-2021 Red Hat, Inc.
```
nickboldt: please update to 2022 as this is a new file and it's 2022.

.github/workflows/release.yml
```diff
@@ -0,0 +1,118 @@
+
+# This Workflow performs a full release of Che Code
```
nickboldt, line 10: [Task as described](https://github.com/eclipse/che/issues/21552#issuecomment-1238161715) was:
"rename image-publish-insiders.yaml to image-publish.yaml and have it handle BOTH main branch -> insiders and next, and add code for handling 7.yy.x -> 7.yy.z releases"

.github/workflows/release.yml
```diff
@@ -0,0 +1,118 @@
...
+      - name: Create success MM message
+        run: |
+          echo "{\"text\":\":white_check_mark: Che Code ${{ github.event.inputs.version }} has been released: https://quay.io/che-incubator/che-code:${{ github.event.inputs.version }}\"}" > mattermost.json
+      - name: Send MM message
+        if: ${{ success() }} || ${{ failure() }}
+        uses: mattermost/action-mattermost-notify@1.1.0
+        env:
+          MATTERMOST_WEBHOOK_URL: ${{ secrets.MATTERMOST_WEBHOOK_URL }}
+          MATTERMOST_CHANNEL: eclipse-che-releases
+          MATTERMOST_USERNAME: che-bot
```
nickboldt: this is an incubating repo, so you probably want the che incubator bot.

.github/workflows/release.yml
nickboldt: but I suppose in other repos we have a release.yml and a next-build.yml...

https://github.com/che-incubator/kubernetes-image-puller-operator/tree/main/.github/workflows
https://github.com/che-incubator/kubernetes-image-puller/tree/main/.github/workflows
https://github.com/che-incubator/chectl/tree/main/.github/workflows

nickboldt: can you merge the functionality of  image-publish-insiders.yaml into the new release.yml so we don't have nearly identical duped files? or @azatsarynnyy are you OK with having a release.yml and a  image-publish-insiders.yaml ?

.github/workflows/image-publish.yml
```diff
@@ -64,10 +74,14 @@ jobs:
       - name: Docker Build and Push
         run: |
           SHORT_SHA1=$(git rev-parse --short=7 HEAD)
-          docker buildx build --platform linux/amd64 -f build/dockerfiles/assembly.Dockerfile --push -t quay.io/che-incubator/che-code:insiders -t quay.io/che-incubator/che-code:next -t quay.io/che-incubator/che-code:insiders-${SHORT_SHA1} .
+          if [[ ${{ env.BRANCH_NAME }} == main ]]; then
+            docker buildx build --platform linux/amd64 -f build/dockerfiles/assembly.Dockerfile --push -t quay.io/che-incubator/che-code:insiders -t quay.io/che-incubator/che-code:next -t quay.io/che-incubator/che-code:insiders-${SHORT_SHA1} .
+          else if [[ ${{ env.BRANCH_NAME }} =~ ^7\.[0-9]+\.[0-9]+$ ]]; then
+            docker buildx build --platform linux/amd64 -f build/dockerfiles/assembly.Dockerfile --push -t quay.io/che-incubator/che-code:${{env.BRANCH_NAME}}  .
```
nickboldt, line 77: need a `fi` to close this if block

.github/workflows/image-publish.yml
```diff
@@ -11,10 +11,12 @@ name: image-publish-insiders
```
nickboldt, line 11: please fix the name of the workflow as it's not just insiders.

.github/workflows/image-publish.yml
```diff
@@ -11,10 +11,12 @@ name: image-publish-insiders
 
 on:
   push:
-    branches: [ main ]
+    branches: 
+      - main
+    tags:
```
nickboldt, line 16: should this fire on changes to a 7.yy.x branch? Or should there be a workflow_dispatch so we can trigger it from the che-release repo as we do for others ? https://github.com/eclipse-che/che-release/blob/main/make-release.sh#L99-L107

.github/workflows/image-publish.yml
```diff
@@ -81,3 +96,18 @@ jobs:
         run: |
           SHORT_SHA1=$(git rev-parse --short=7 HEAD)
           docker buildx build --platform linux/amd64 -f build/dockerfiles/dev.Dockerfile --push -t quay.io/che-incubator/che-code-dev:insiders -t quay.io/che-incubator/che-code-dev:next -t quay.io/che-incubator/che-code-dev:insiders-${SHORT_SHA1} .
```
nickboldt, line 117: you've already done this line above on line 79.

.github/workflows/image-publish.yml
```diff
@@ -64,10 +75,14 @@ jobs:
...
           SHORT_SHA1=$(git rev-parse --short=7 HEAD)
-          docker buildx build --platform linux/amd64 -f build/dockerfiles/assembly.Dockerfile --push -t quay.io/che-incubator/che-code:insiders -t quay.io/che-incubator/che-code:next -t quay.io/che-incubator/che-code:insiders-${SHORT_SHA1} .
+          if [[ ${{ env.BRANCH_NAME }} == main ]]; then
+            docker buildx build --platform linux/amd64 -f build/dockerfiles/assembly.Dockerfile --push -t quay.io/che-incubator/che-code:insiders -t quay.io/che-incubator/che-code:next -t quay.io/che-incubator/che-code:insiders-${SHORT_SHA1} .
+          else if [[ ${{ env.BRANCH_NAME }} =~ ^7\.[0-9]+\.[0-9]+$ ]]; then
+            docker buildx build --platform linux/amd64 -f build/dockerfiles/assembly.Dockerfile --push -t quay.io/che-incubator/che-code:${{env.BRANCH_NAME}}  . 
   dev:
     name: dev
     runs-on: ubuntu-20.04
+    if: github.ref == 'refs/heads/main'
```
nickboldt, line 104: So we only create MM messages for insiders, not for 7.yy.x releases?

.github/workflows/image-publish.yml
nickboldt: oops, I see this is for che-code-dev, not che-code. Does this need to be done BEFORE the che-code image?

.github/workflows/image-publish.yml
```diff
@@ -81,3 +96,18 @@ jobs:
         run: |
           SHORT_SHA1=$(git rev-parse --short=7 HEAD)
           docker buildx build --platform linux/amd64 -f build/dockerfiles/dev.Dockerfile --push -t quay.io/che-incubator/che-code-dev:insiders -t quay.io/che-incubator/che-code-dev:next -t quay.io/che-incubator/che-code-dev:insiders-${SHORT_SHA1} .
+      - name: Create failure MM message
+        if: ${{ failure() }} || {{ github.ref == 'refs/heads/main'}}
+        run: |
+          echo "{\"text\":\":no_entry_sign: Che Code ${{ github.event.inputs.version }} release has failed: https://github.com/che-incubator/che-code/actions/workflows/release.yml\"}" > mattermost.json
```
nickboldt: `github.event.inputs.version` is not defined unless you add a `workflow_dispatch` with a `version` variable.

make-release.sh
```diff
@@ -121,18 +121,3 @@ if [[ $TAG_RELEASE -eq 1 ]]; then
   git tag "${VERSION}"
   git push origin "${VERSION}"
 fi
-
-# now update ${BASEBRANCH} to the new version
```
nickboldt: why remove this? Don't we need a version in main and in 7.yy.x branches ?

.github/workflows/release.yml
```diff
@@ -0,0 +1,104 @@
...
+    runs-on: ubuntu-20.04
+    steps:
+      -
+        name: "Checkout Che Code source code"
+        uses: actions/checkout@v2
+      -
+        name: "Setup Node"
+        uses: actions/setup-node@v2
+        with:
+          node-version: "14"
```
nickboldt: can we use node 16? 14 is nearly EOL. https://endoflife.date/nodejs

.github/workflows/image-publish.yml
```diff
@@ -81,3 +96,18 @@ jobs:
         run: |
           SHORT_SHA1=$(git rev-parse --short=7 HEAD)
           docker buildx build --platform linux/amd64 -f build/dockerfiles/dev.Dockerfile --push -t quay.io/che-incubator/che-code-dev:insiders -t quay.io/che-incubator/che-code-dev:next -t quay.io/che-incubator/che-code-dev:insiders-${SHORT_SHA1} .
+      - name: Create failure MM message
```
nickboldt: this should happen for a tag build, not for insiders/next builds.

.github/workflows/release.yml
nickboldt: Sounds like Artem is OK with two files, so let's go with what you have right now.

nickboldt (changes requested):
small tweaks needed but otherwise +1

.github/workflows/release.yml
mkuznyetsov: we use che-bot for mattermost

.github/workflows/release.yml
azatsarynnyy: I'm OK with both variants. If two files are more clear from your PoV - let's use that approach.

.github/workflows/image-publish.yml
```diff
@@ -77,8 +76,24 @@ jobs:
           SHORT_SHA1=$(git rev-parse --short=7 HEAD)
           if [[ ${{ env.BRANCH_NAME }} == main ]]; then
             docker buildx build --platform linux/amd64 -f build/dockerfiles/assembly.Dockerfile --push -t quay.io/che-incubator/che-code:insiders -t quay.io/che-incubator/che-code:next -t quay.io/che-incubator/che-code:insiders-${SHORT_SHA1} .
-          else if [[ ${{ env.BRANCH_NAME }} =~ ^7\.[0-9]+\.[0-9]+$ ]]; then
+          elif [[ ${{ env.BRANCH_NAME }} =~ ^7\.[0-9]+\.[0-9]+$ ]]; then
             docker buildx build --platform linux/amd64 -f build/dockerfiles/assembly.Dockerfile --push -t quay.io/che-incubator/che-code:${{env.BRANCH_NAME}}  . 
+          fi
+      - name: Create failure MM message
+        if: ${{ failure() }} || {{ github.ref != 'refs/heads/main'}}
```
nickboldt: I think you mean if failure AND ref != main

.github/workflows/image-publish.yml
```diff
@@ -77,8 +76,24 @@ jobs:
...
+        run: |
+          echo "{\"text\":\":no_entry_sign: Che Code ${{ env.BRANCH_NAME }} release has failed: https://github.com/che-incubator/che-code/actions/workflows/release.yml\"}" > mattermost.json
+      - name: Create success MM message
+        if: ${{ github.ref != 'refs/heads/main'}}
+        run: |
+          echo "{\"text\":\":white_check_mark: Che Code ${{ env.BRANCH_NAME }} has been released: https://quay.io/che-incubator/che-code:${{ env.BRANCH_NAME }}\"}" > mattermost.json
+      - name: Send MM message
+        if: ${{ success() }} || ${{ failure() }} || ${{ github.ref != 'refs/heads/main'}}
```
nickboldt: I think you mean if (success or failure) AND ref != main...

.github/workflows/image-publish.yml
```diff
@@ -77,8 +76,24 @@ jobs:
           SHORT_SHA1=$(git rev-parse --short=7 HEAD)
           if [[ ${{ env.BRANCH_NAME }} == main ]]; then
             docker buildx build --platform linux/amd64 -f build/dockerfiles/assembly.Dockerfile --push -t quay.io/che-incubator/che-code:insiders -t quay.io/che-incubator/che-code:next -t quay.io/che-incubator/che-code:insiders-${SHORT_SHA1} .
-          else if [[ ${{ env.BRANCH_NAME }} =~ ^7\.[0-9]+\.[0-9]+$ ]]; then
+          elif [[ ${{ env.BRANCH_NAME }} =~ ^7\.[0-9]+\.[0-9]+$ ]]; then
             docker buildx build --platform linux/amd64 -f build/dockerfiles/assembly.Dockerfile --push -t quay.io/che-incubator/che-code:${{env.BRANCH_NAME}}  . 
+          fi
+      - name: Create failure MM message
```
nickboldt, line 86: can you wrap this whole section in a single if ref != main instead of having to check for that 3 times? If so it would make your OR+AND block below simler to implement. 

Maybe move it out of `jobs.build.steps` and into `jobs.announce.steps`? 

You did the exact thing I'm thinking of in jobs.dev:

https://github.com/che-incubator/che-code/blob/79d2745bb4b31236271b7ed26e85fb64c4852031/.github/workflows/image-publish.yml#L100

.github/workflows/image-publish.yml
```diff
@@ -77,8 +76,24 @@ jobs:
           SHORT_SHA1=$(git rev-parse --short=7 HEAD)
           if [[ ${{ env.BRANCH_NAME }} == main ]]; then
```
nickboldt, line 77: If you're setting BRANCH_NAME to only equal `main` instead of `refs/heads/main` then your if-check on line 50 is going to fail because you removed `refs/heads/` on line 49

.github/workflows/image-publish.yml
```diff
@@ -7,14 +7,13 @@
...
-name: image-publish-insiders
+name: image-publish
 
 on:
   push:
     branches: 
       - main
-    tags:
-      - '^7\.[0-9]+\.[0-9]+$'
+  workflow_dispatch:
```
nickboldt: I think you need to set some variables here or this block will fail. In particular, if we're triggering this by hand, don't we have to set a branch/tag value to release?

nickboldt: see this action:

* https://github.com/redhat-developer/devspaces/blob/devspaces-3-rhel-8/.github/workflows/plugin-registry-build-publish-content-gh-pages.yaml#L21-L26 support user input override for default
* https://github.com/redhat-developer/devspaces/blob/devspaces-3-rhel-8/.github/workflows/plugin-registry-build-publish-content-gh-pages.yaml#L37-L55 set an env var depending on if we have user input or the default 
* https://github.com/redhat-developer/devspaces/blob/devspaces-3-rhel-8/.github/workflows/plugin-registry-build-publish-content-gh-pages.yaml#L65-L68 use the value defined by default or user input

.github/workflows/image-publish.yml
```diff
@@ -42,6 +43,15 @@ jobs:
     needs: build
     runs-on: ubuntu-20.04
     steps:
+      - name: Set variables:
+        run: |
+          BRANCH_NAME=${{ github.ref }}
+          BRANCH_NAME=${BRANCH_NAME##*/}
+          if [[ $BRANCH_NAME == 'refs/heads/main' ]]; then
```
nickboldt: If you're setting BRANCH_NAME to only equal `main` instead of `refs/heads/main` then your if-check on line 50 is going to fail because you removed `refs/heads/` on line 49

.github/workflows/image-publish.yml
```diff
@@ -64,10 +74,30 @@ jobs:
       - name: Docker Build and Push
         run: |
           SHORT_SHA1=$(git rev-parse --short=7 HEAD)
-          docker buildx build --platform linux/amd64 -f build/dockerfiles/assembly.Dockerfile --push -t quay.io/che-incubator/che-code:insiders -t quay.io/che-incubator/che-code:next -t quay.io/che-incubator/che-code:insiders-${SHORT_SHA1} .
+          if [[ ${{ env.BRANCH_NAME }} == main ]]; then
```
nickboldt, line 74: You set `env.BRANCH_NAME` to "insiders" on line 51, so this check will never pass.

.github/workflows/image-publish.yml
```diff
@@ -79,16 +80,21 @@ jobs:
           elif [[ ${{ env.BRANCH_NAME }} =~ ^7\.[0-9]+\.[0-9]+$ ]]; then
             docker buildx build --platform linux/amd64 -f build/dockerfiles/assembly.Dockerfile --push -t quay.io/che-incubator/che-code:${{env.BRANCH_NAME}}  . 
           fi
+  release-finish:
+    name: release-finish
+    needs: [build, assembly]
+    runs-on: ubuntu-20.04
+    if: (success() || failure()) && github.ref != 'refs/heads/main'
```
nickboldt, line 84: is there another status beside success and failure? if not, then can we simplify and do:
```
    if: github.ref != 'refs/heads/main'
```

.github/workflows/image-publish.yml
```diff
@@ -47,8 +48,8 @@ jobs:
         run: |
           BRANCH_NAME=${{ github.ref }}
           BRANCH_NAME=${BRANCH_NAME##*/}
-          if [[ $BRANCH_NAME == 'refs/heads/main' ]]; then
-            echo "BRANCH_NAME=insiders" >> $GITHUB_ENV
+          if [[ $BRANCH_NAME == 'main' ]]; then
+            echo "BRANCH_NAME=main" >> $GITHUB_ENV
```
nickboldt: you could simplify this to:
```
if [[ $BRANCH_NAME == 'main' ]] || [[ $BRANCH_NAME =~ ^7\.[0-9]+\.[0-9]+$ ]]; then
  echo "BRANCH_NAME=$BRANCH_NAME >> $GITHUB_ENV
fi
```

But since you're not trapping for other requests and failing the build for non-compliant values... you could also just do this:

```
BRANCH_NAME=${{ github.ref }}
echo "BRANCH_NAME=${BRANCH_NAME##*/}" >> $GITHUB_ENV
```

and no if/then/fi block needed.

.github/workflows/image-publish.yml
```diff
@@ -1,20 +1,22 @@
 #
-# Copyright (c) 2021 Red Hat, Inc.
+# Copyright (c) 2022 Red Hat, Inc.
```
nickboldt: this should technically be 2021-2022, since the renamed file already existed.

make-release.sh
```diff
@@ -0,0 +1,139 @@
+#!/bin/bash
+#
+# Copyright (c) 2020-2021 Red Hat, Inc.
```
nickboldt: new file: copyright starts at 2022.

.github/workflows/image-publish.yml
mkuznyetsov: I dont need this to be run on cancelled builds

.github/workflows/image-publish.yml
mkuznyetsov: thanks!

azatsarynnyy (approved):
LGTM

.github/workflows/image-publish.yml
```diff
@@ -64,10 +71,35 @@ jobs:
       - name: Docker Build and Push
         run: |
           SHORT_SHA1=$(git rev-parse --short=7 HEAD)
-          docker buildx build --platform linux/amd64 -f build/dockerfiles/assembly.Dockerfile --push -t quay.io/che-incubator/che-code:insiders -t quay.io/che-incubator/che-code:next -t quay.io/che-incubator/che-code:insiders-${SHORT_SHA1} .
+          if [[ ${{ env.BRANCH_NAME }} == main ]]; then
+            docker buildx build --platform linux/amd64 -f build/dockerfiles/assembly.Dockerfile --push -t quay.io/che-incubator/che-code:insiders -t quay.io/che-incubator/che-code:next -t quay.io/che-incubator/che-code:insiders-${SHORT_SHA1} .
+          elif [[ ${{ env.BRANCH_NAME }} =~ ^7\.[0-9]+\.[0-9]+$ ]]; then
+            docker buildx build --platform linux/amd64 -f build/dockerfiles/assembly.Dockerfile --push -t quay.io/che-incubator/che-code:${{env.BRANCH_NAME}}  .
```
nickboldt, line 77: I guess technically this is either a branch or a tag, but that's fine as Git often uses "branch" for a "branch-or-tag-or-other-identifier". +1.

.github/workflows/image-publish.yml
```diff
@@ -64,10 +71,35 @@ jobs:
...
+          if [[ ${{ env.BRANCH_NAME }} == main ]]; then
+            docker buildx build --platform linux/amd64 -f build/dockerfiles/assembly.Dockerfile --push -t quay.io/che-incubator/che-code:insiders -t quay.io/che-incubator/che-code:next -t quay.io/che-incubator/che-code:insiders-${SHORT_SHA1} .
+          elif [[ ${{ env.BRANCH_NAME }} =~ ^7\.[0-9]+\.[0-9]+$ ]]; then
+            docker buildx build --platform linux/amd64 -f build/dockerfiles/assembly.Dockerfile --push -t quay.io/che-incubator/che-code:${{env.BRANCH_NAME}}  . 
+          fi
+  release-finish:
+    name: release-finish
+    needs: [build, assembly]
+    runs-on: ubuntu-20.04
+    if: (success() || failure()) && github.ref != 'refs/heads/main'
```
nickboldt, line 84: I don't know why you have "if success or failure" here... are there other statuses to check like aborted/cancelled? Do we care? Is the goal the always create the MM message? If so I think you only need:
```
    if: github.ref != 'refs/heads/main'
```

.github/workflows/image-publish.yml
```diff
@@ -64,10 +71,35 @@ jobs:
...
+    steps:
+      - name: Create failure MM message
+        if: failure()
+        run: |
+          echo "{\"text\":\":no_entry_sign: Che Code ${{ env.BRANCH_NAME }} release has failed: https://github.com/che-incubator/che-code/actions/workflows/release.yml\"}" > mattermost.json
+      - name: Create success MM message
+        run: |
+          echo "{\"text\":\":white_check_mark: Che Code ${{ env.BRANCH_NAME }} has been released: https://quay.io/che-incubator/che-code:${{ env.BRANCH_NAME }}\"}" > mattermost.json
+      - name: Send MM message
+        if: success() || failure()
```
nickboldt, line 95: once again, does this `if` check do anything?

.github/workflows/image-publish.yml
```diff
@@ -64,10 +71,35 @@ jobs:
...
+        uses: mattermost/action-mattermost-notify@1.1.0
+        env:
+          MATTERMOST_WEBHOOK_URL: ${{ secrets.MATTERMOST_WEBHOOK_URL }}
+          MATTERMOST_CHANNEL: eclipse-che-releases
+          MATTERMOST_USERNAME: che-bot
   dev:
     name: dev
```
nickboldt, line 102: might want to rename this to "build-dev" for clarity; also @azatsarynnyy does the dev image need to be built BEFORE the other images? Is it like with Che-theia there che-theia depends on che-theia-dev?

make-release.sh
```diff
@@ -0,0 +1,139 @@
...
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
+# Release process automation script.
+# Used to create branch/tag, update VERSION files
+# and and trigger release by force pushing changes to the release branch
```
nickboldt: this comment is not correct. 

How about
```
Used to create a branch update the version stored in package.json, and push a tag
Once the tag is created, image-publish.yml will be triggered to build the container image.
```

.github/workflows/image-publish.yml
azatsarynnyy: no, Che-Code image doesn't depend on the dev image
so, the build order doesn't matter

nickboldt (approved):

make-release.sh
mkuznyetsov: thx

PR #120, opened 2022-09-27
Add an ability to use custom Node.js for running VS Code

RomanNikitenko:
Issue: https://github.com/eclipse/che/issues/21629, https://issues.redhat.com/browse/CRW-3261

At the moment for running VS Code `che-code` uses `Node.js 16.14.0` that requires `OpenSSL 1.1.1`.
There is no problem when `UBI 8` based image is used in a devfile (`UBI 8` image uses 1.1.1 version of `openssl`). But it's impossible to start a workspace when `UBI 9` based custom image is used (`UBI 9` image uses 3.0.1 version of `openssl`). Please see screenshots [here](https://github.com/eclipse/che/issues/21629#issuecomment-1250917204).
   
The changes allow a user to use an `UBI 9` based custom image in a devfile. The requirements are:
- install `Node.js 16` which statically links against `OpenSSL 3`
- provide `VSCODE_NODEJS_RUNTIME_DIR` environment variable with path to the installed node

image

At the same time within a custom image a user can use any version of Node.js as default - I don't see here a problem.

node_versions


build/scripts/entrypoint-volume.sh
```diff
@@ -72,5 +72,11 @@ if [ -f /tmp/che/secret/ca.crt ]; then
   export NODE_EXTRA_CA_CERTS=/tmp/che/secret/ca.crt
 fi
 
+if [ -z "$NODEJS_FOR_VSCODE_DIR" ]; then
```
RomanNikitenko: maybe we need better variable name here...

azatsarynnyy: maybe `VSCODE_NODEJS_RUNTIME_DIR` ?

azatsarynnyy (approved):

PR #162, opened 2023-01-07
chore: Update ubuntu to 22.04 and actions to node 16 versions

SDawley:
Signed-off-by: sdawley <sdawley@redhat.com>

### What does this PR do?
Updates Ubuntu to 22.04 and all actions to the latest version using node 16

https://github.com/eclipse/che/issues/21763

### DIFF
--- a/.github/workflows/try-in-web-ide.yaml
+++ b/.github/workflows/try-in-web-ide.yaml
@@ -18,16 +18,16 @@ on:
 
 jobs:
   add-link:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - name: Add DevSandbox link
-        uses: redhat-actions/try-in-web-ide@v1
+        uses: redhat-actions/try-in-web-ide@v1.2
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           add_comment: true
           add_status: false
       - name: Add Dogfooding link
-        uses: redhat-actions/try-in-web-ide@v1
+        uses: redhat-actions/try-in-web-ide@v1.2
         with:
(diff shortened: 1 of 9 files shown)


.github/workflows/try-in-web-ide.yaml
```diff
@@ -18,16 +18,16 @@ on:
...
       - name: Add DevSandbox link
-        uses: redhat-actions/try-in-web-ide@v1
+        uses: redhat-actions/try-in-web-ide@v1.2
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           add_comment: true
           add_status: false
       - name: Add Dogfooding link
-        uses: redhat-actions/try-in-web-ide@v1
+        uses: redhat-actions/try-in-web-ide@v1.2
```
nickboldt, line 30: why 1.2 here? are we avoiding 1.3+ ?

azatsarynnyy: Currently, 1.2 is the latest - https://github.com/redhat-actions/try-in-web-ide/releases

azatsarynnyy (approved):
Thanks @SDawley for checking all the actions!

.github/workflows/try-in-web-ide.yaml
azatsarynnyy: Currently, 1.2 is the latest - https://github.com/redhat-actions/try-in-web-ide/releases

nickboldt: yes, but @1 == @1.x so this change is actually PINNING to 1.2 rather than using 1.latest

PR #163, opened 2023-01-10
feat: Handle disconnection according to the dev workspace status

RomanNikitenko:

### What does this PR do?
Handle disconnection according to the dev workspace status:
- Run timeout
image

- Inactivity timeout
image

- An error
image

- Workspace is not running
image

 Within testing I detected the following unusual use case:
- the disconnection happened
- the API responds that the workspace is running
- dashboard displays the workspace as running
- but the workspace is not usable

I've added some logic to propose reloading the window for such use case - it fixes the disconnection. See the behaviour on the video.

https://user-images.githubusercontent.com/5676062/212892918-088eb7c7-002b-442f-99dd-a427037ab099.mp4

    

https://github.com/eclipse/che/issues/21610

### How to test this PR?
It's possible to change the corresponding properties to test use cases related to timeouts:
<details>
<summary>See screenshot here </summary>
timeouts

</details>

#### Run timeout
1. Set the workspace run timeout (`checluster.spec.devEnvironments.secondsOfRunBeforeIdling`) to 120, for example.
2. Wait about 120 seconds.
3. The corresponding dialog should be displayed.

####  Inactivity timeout
1. Set the workspace inactivity timeout (`checluster.spec.devEnvironments.secondsOfInactivityBeforeIdling`) to 100, for example.
2. Do not interact with the workspace 100 seconds.
3. The corresponding dialog should be displayed.

#### An error
I tested the use case using minikube:
- start a workspace using minikube
- go to your local terminal => `minikube ssh` => `docker ps`
- detect a container with the running `VS Code` and stop it, like: `docker stop d9bd0222b513`
 
docker_ps
 

#### Workspace is not running
The use case is related to the any case when 
- a workspace is not running
- we can not detect an error, timeout or any reason of the stopped workspace

You can just stop a workspace using dashboard to test it.

### DIFF
--- a/.rebase/replace/code/src/vs/workbench/contrib/remote/browser/remote.ts.json
+++ b/.rebase/replace/code/src/vs/workbench/contrib/remote/browser/remote.ts.json
@@ -1,0 +1,34 @@
+[
+    {
+        "from": "import { getVirtualWorkspaceLocation } from 'vs/platform/workspace/common/virtualWorkspace';",
+        "by": "import { getVirtualWorkspaceLocation } from 'vs/platform/workspace/common/virtualWorkspace';\\\nimport { IRequestService } from 'vs/platform/request/common/request';\\\nimport { CheDisconnectionHandler } from 'vs/workbench/contrib/remote/browser/che/remote';\\\nimport { INotificationService } from 'vs/platform/notification/common/notification';\\\nimport { IEnvironmentVariableService } from 'vs/workbench/contrib/terminal/common/environmentVariable';"
+    },
+    {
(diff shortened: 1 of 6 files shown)


azatsarynnyy (commented):
To test the situation when the user might receive this message
image

I've just sent a SIGKILL to the VS Code process from its terminal (`kill -9 pid`). In this case, DevWorkspace is still running, but VS Code is not.

azatsarynnyy (approved):
I tested different use cases - everything works as described :+1: 

Thanks for updating the rebase rules.

PR #178, opened 2023-02-09
fix: Fix VS Code build for s390x platform

RomanNikitenko:

`@vscode/l10n-dev` dependency was upgraded on the VS Code side.
The new version of `@vscode/l10n-dev` brings transitive dependency `web-tree-sitter`.
As far as I understand `web-tree-sitter` is incompatible to `s390x` platform. 
As result we have a failing build on the `s390x` platform:

image

I see a clear message at attempt to install the lib:

image

I played with the lib on the `s390x` platform, debugged it to find how we can fix it.
Unfortunately I didn't find a way how to get `web-tree-sitter` working for the `s390x` platform, maybe we need more time for an investigation.  

### What does this PR do?
My propose is:
- we use `0.0.18` version of `@vscode/l10n-dev`
- it ^ doesn't contain `web-tree-sitter` transitive dependency
- it has a difference in API, but I provided an adaptation for that difference
- the approach allows to keep the localization functionality without breaking changes for the `s390x` platform 

https://issues.redhat.com/browse/CRW-3914

### How to test this PR?

- I tried to build `che-code` on the `s390x` platform - it was successful
- after merging this PR we can check the corresponding jobs on the jenkins
- the changes are related to localization tooling for Visual Studio Code. At the moment this functionality doesn't work for Che Code (please see https://github.com/eclipse/che/issues/21963), so there is no way to check it.


nickboldt (approved):
+1 for this idea

PR #187, opened 2023-03-06
[WIP] test loading webview resources from local container

vitaliy-guliy:

### DIFF
--- a/code/src/vs/workbench/contrib/webview/common/webview.ts
+++ b/code/src/vs/workbench/contrib/webview/common/webview.ts
@@ -3,6 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+// import { env } from 'process';
 import { CharCode } from 'vs/base/common/charCode';
 import { Schemas } from 'vs/base/common/network';
 import { URI } from 'vs/base/common/uri';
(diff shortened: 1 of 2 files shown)


miboettc:
@vitaliy-guliy : Many thanks for trying to solve the webview issue! Recently I discovered that another downstream project of VS Code had faced similar problems. What they patched to make webviews also work locally is documented here: https://github.com/coder/code-server/blob/main/patches/webview.diff 

They also have a patch for ServiceWorkers (https://github.com/coder/code-server/blob/main/patches/service-worker.diff) but here I'm less sure if they address the webview problem.

vitaliy-guliy:
> @vitaliy-guliy : Many thanks for trying to solve the webview issue! Recently I discovered that another downstream project of VS Code had faced similar problems. What they patched to make webviews also work locally is documented here: https://github.com/coder/code-server/blob/main/patches/webview.diff
> 
> They also have a patch for ServiceWorkers (https://github.com/coder/code-server/blob/main/patches/service-worker.diff) but here I'm less sure if they address the webview problem.

Thanks a lot for pointing me on the solution they did!
Yes, it's the similar issue. But instead of getting the resources from current location [we defined a different endpoint](https://github.com/eclipse-che/che-plugin-registry/pull/1600) to the same che-code backend. It is necessary to make the webview content isolated from the main frame.

PR #191, opened 2023-03-17
feat: Add 'Open Cluster Console' command

RomanNikitenko:
depends on https://github.com/eclipse-che/che-dashboard/pull/758

### What does this PR do?
- adds `Open OpenShift Console` command to the Command Palette
- adds `Open OpenShift Console` action to the `Eclipse Che` status bar actions

https://github.com/eclipse/che/issues/21822

### How to test this PR?
Please see a short video:

https://user-images.githubusercontent.com/5676062/226611639-b3b971b1-f8c8-439d-80c5-1f7dfd52667c.mp4

### DIFF
--- a/code/extensions/che-remote/package.nls.json
+++ b/code/extensions/che-remote/package.nls.json
@@ -3,7 +3,7 @@
 	"description": "Provides remoteIndicator commands",
 	"openDocumentationCommand": "Eclipse Che: Open Documentation",
 	"openDashboardCommand": "Eclipse Che: Open Dashboard",
-	"openClusterConsoleCommand": "Eclipse Che: Open Cluster Console",
+	"openClusterConsoleCommand": "Eclipse Che: Open OpenShift Console",
 	"stopWorkspaceCommand": "Eclipse Che: Stop Workspace",
 	"restartWorkspaceCommand": "Eclipse Che: Restart Workspace",
 	"restartWorkspaceFromLocalDevfileCommand": "Eclipse Che: Restart Workspace from Local Devfile"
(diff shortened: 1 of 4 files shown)


code/extensions/che-remote/package.nls.json
```diff
@@ -3,6 +3,7 @@
 	"description": "Provides remoteIndicator commands",
 	"openDocumentationCommand": "Eclipse Che: Open Documentation",
 	"openDashboardCommand": "Eclipse Che: Open Dashboard",
+	"openClusterConsoleCommand": "Eclipse Che: Open Cluster Console",
```
l0rd:
```suggestion
	"openClusterConsoleCommand": "Eclipse Che: Open OpenShift Console",
```

RomanNikitenko: Can we rely that it's always `OpenShift`?
or
it could have another value?

I'm asking because I see that [on the dashboard side there is specific env variable](https://github.com/eclipse-che/che-dashboard/blob/b229ff4598d2af3990a8daa0cd71bdf061ec740d/packages/dashboard-backend/src/routes/api/clusterInfo.ts#L28) which contains a title. The dashboard reads that env variable and provides the title based on the given value.

About the `che-code` side - we could use that env variable as well **if it's really required** (so, title can have different value). But then the more complex solution needed:
- VS Code doesn't allow dynamic titles for commands provided by an extension (the names are defined in the `package.json` file of the extension)
- so, some part should be provided on the VS Code side + some part on an extension side

l0rd: There are other Kubernetes GUIs but they are not named "Console" (c.f. [the Kubernetes Dashboard](https://kubernetes.io/docs/tasks/access-application-cluster/web-ui-dashboard/)) so "Cluster Console" is not a generic enough name and something like "Kubernetes GUI" is generic enough but it may be overkill considered that we retrieve the URL if it's OpenShift only.

That said we should:
- hide the menu item / command if VS Code is **not** running on OpenShift 
- use the "OpenShift Console" label

azatsarynnyy (approved):
:+1:

PR #192, opened 2023-03-24
fix: Filter out the editor-provided commands

### LINKED ISSUE #22098: Don't translate the editor-provided Devfile commands to VS Code Tasks

I noticed one small side-effect of adding [the command of `exec` type in the editor definition](https://github.com/eclipse-che/che-plugin-registry/pull/1566/files#diff-4526f5cc968bddb01b7aaa7b95bbf3be602a7cfe9b7a69230ecbef43751732b4R347-R350):
Now, it's listed among all the VS Code Tasks of the `devfile` type:
image

This command is needed just for running the VS Code. It's useless for the user, as a VS Code Task.

The solution can be filtering out the `exec` Devfile commands provided by the editor definition.
We can filter them by the `controller.devfile.io/imported-by: editor` attribute in the flattened Devfile.

_No response_

### Additional context

https://github.com/eclipse/che/issues/21879#issuecomment-1481635823

azatsarynnyy:
### What does this PR do?
Filters out the editor-provided Devfile commands as they are not expected to be run by the user.

Closes https://github.com/eclipse/che/issues/22098

### How to test this PR?
1. Start a new workspace from the test project https://github.com/azatsarynnyy/java-spring-petclinic/tree/hide-editor-exec-cmds. Its `.che/che-editor.yaml` contains the `init-che-code-command` exec command.
2. Verify that there is no editor-specific command among the `devfile`-type Tasks.

### DIFF
--- a/code/extensions/che-commands/src/taskProvider.ts
+++ b/code/extensions/che-commands/src/taskProvider.ts
@@ -37,6 +37,7 @@ export class DevfileTaskProvider implements vscode.TaskProvider {
 
 		const cheTasks: vscode.Task[] = devfileCommands!
 			.filter(command => command.exec?.commandLine)
+			.filter(command => !command.attributes || (command.attributes as any)['controller.devfile.io/imported-by'] !== 'editor')
 			.map(command => this.createCheTask(command.exec?.label || command.id, command.exec?.commandLine!, command.exec?.workingDir || '${PROJECT_SOURCE}', command.exec?.component!));
 		return cheTasks;
 	}


RomanNikitenko (approved):

code/extensions/che-commands/src/taskProvider.ts
```diff
@@ -37,6 +37,7 @@ export class DevfileTaskProvider implements vscode.TaskProvider {
 
 		const cheTasks: vscode.Task[] = devfileCommands!
 			.filter(command => command.exec?.commandLine)
+			.filter(command => !command.attributes || (command.attributes as any)['controller.devfile.io/imported-by'] !== 'editor')
```
amisevsk, line 40: I'm worried that this is a somewhat fragile approach; the attribute `controller.devfile.io/imported-by: editor` comes from the name of the plugin/contribution that imported it, so this works only as long as devfiles specify the name `editor` for the plugin. Currently, as far as I can tell, this is done in the devfile registry, [here](https://github.com/eclipse-che/che-devfile-registry/blob/25f630b689d3fbc90e23a798a15e2ef3c9a17002/tools/devworkspace-generator/src/generate.ts#L77-L82), which results in devworkspaces that contain
```yaml
  contributions:
    - name: editor # <--- source for the attribute
      kubernetes:
        name: <workspace-specific-name>
```

This may also pose a problem in DevWorkspaces created in an earlier version of Che, as previously the editor contribution was named differently (e.g. in my sandbox workspaces, I have 
```yaml
    commands:
    - apply:
        component: che-code-injector
      attributes:
        controller.devfile.io/imported-by: che-code-golang-example
      id: init-container-command
```


Another approach, that would be safer but more complicated, would be to filter commands that are tied to event bindings in the flattened DevWorkspace. In other words, we could look at
* `.events.prestart[]`
* `.events.poststart[]`
* `.events.prestop[]`

and filter commands if their id appears in any of those lists. Going off the above example, my (flattened) DevWorkspace has the following events:
```yaml
    events:
      postStart:
      - init-che-code-command
      preStart:
      - init-container-command
```

alternatively, we could filter any commands that were imported by plugins/contributions (i.e. any commands that have the `controller.devfile.io/imported-by` attribute, regardless of value), though this would exclude any commands that came from any contribution. Currently these are only used for editors, so maybe this isn't a problem?

azatsarynnyy: Hi @amisevsk 
Thanks a lot for noticing this!
Filtering any commands that were imported by plugins/contributions sounds perfect. It's simple and does what we need. I fixed it in https://github.com/che-incubator/che-code/pull/194

PR #194, opened 2023-03-28
fix: Filter out che commands imported by plugins/contributions

azatsarynnyy:
### What does this PR do?
Filters out che commands imported by plugins/contributions as they are not expected to be run by the user.

https://github.com/eclipse/che/issues/22098
implements the suggestion https://github.com/che-incubator/che-code/pull/192#pullrequestreview-1359995830

### How to test this PR?
1. Start a new workspace from the test project https://github.com/azatsarynnyy/java-spring-petclinic/tree/filter-cmds. The project is configured to use the editor image built from this PR.
2. Verify that there is no `init-che-code-command` command among the `devfile`-type Tasks.

### DIFF
--- a/code/extensions/che-commands/src/taskProvider.ts
+++ b/code/extensions/che-commands/src/taskProvider.ts
@@ -37,7 +37,7 @@ export class DevfileTaskProvider implements vscode.TaskProvider {
 
 		const cheTasks: vscode.Task[] = devfileCommands!
 			.filter(command => command.exec?.commandLine)
-			.filter(command => !command.attributes || (command.attributes as any)['controller.devfile.io/imported-by'] !== 'editor')
+			.filter(command => !command.attributes || (command.attributes as any)['controller.devfile.io/imported-by'] === undefined)
 			.map(command => this.createCheTask(command.exec?.label || command.id, command.exec?.commandLine!, command.exec?.workingDir || '${PROJECT_SOURCE}', command.exec?.component!));
 		return cheTasks;
 	}


code/extensions/che-commands/src/taskProvider.ts
```diff
@@ -37,7 +37,7 @@ export class DevfileTaskProvider implements vscode.TaskProvider {
 
 		const cheTasks: vscode.Task[] = devfileCommands!
 			.filter(command => command.exec?.commandLine)
-			.filter(command => !command.attributes || (command.attributes as any)['controller.devfile.io/imported-by'] !== 'editor')
+			.filter(command => !command.attributes || (command.attributes as any)['controller.devfile.io/imported-by'] === undefined)
```
amisevsk, line 40: Another case that just occurred to me that we may want to cover -- there's a special case when devfiles use `parent`s. In this case, the command would have the attribute `controller.devfile.io/imported-by: "parent"` (hard-coded, by DWO). Should we include those commands? 

E.g. see the example in this issue: https://github.com/eclipse/che/issues/22103 -- we likely want to include the commands from the [java-maven](https://registry.devfile.io/devfiles/java-maven) parent.

amisevsk (approved):

code/extensions/che-commands/src/taskProvider.ts
azatsarynnyy: Good catch. Thanks @amisevsk !
I've prepared a fix in the draft PR https://github.com/che-incubator/che-code/pull/197/files. Will test it once the dogfoding instance is fixed (https://github.com/eclipse/che/issues/22108) and merge.

PR #197, opened 2023-03-29
fix: Include the commands imported from a parent Devfile as to VS Code Tasks

### LINKED ISSUE #22373: Parent commands are not available in VS Code tasks
### Describe the bug

When a devfile uses a parent and the parent has some commands those are not visible as tasks in VS Code. This is probably due to the fact that the commands have the attribute `controller.devfile.io/imported-by: parent` in the flattened devfile:

```yaml
- attributes:
    controller.devfile.io/imported-by: parent
  exec:
    commandLine: ./mvnw clean package -DskipTests=true
    component: tools
    group:
      isDefault: true
      kind: build
    label: 1. Package
  id: package
```

### Che version

7.71@latest

### Steps to reproduce

Start a workspace using the branch `parent` of this git repository: https://github.com/l0rd/inventory-quarkus/

### Expected behavior

I would like to see the parents commands but I only see the ones coming from the devfile:

image

azatsarynnyy:
### What does this PR do?
When Che-Code translates the Devfile Commands to VS Code Tasks, it ignored the commands defined in a parent Devfile. With this patch, parent Commands are imported now.

addresses the https://github.com/che-incubator/che-code/pull/194#discussion_r1151103256
fixes https://github.com/eclipse/che/issues/22373

### How to test this PR?
1. Spin up a new workspace from the test project https://github.com/azatsarynnyy/inventory-quarkus/tree/with-parent which is configured to use the Che-Code image built from this PR.
2. Open VS Code Tasks list and verify that the commands defined in the parent Devfile are listed as well as the commands defined in the project's Devfile:
image


code/extensions/che-commands/src/taskProvider.ts
```diff
@@ -38,7 +38,7 @@ export class DevfileTaskProvider implements vscode.TaskProvider {
 
 		const cheTasks: vscode.Task[] = devfileCommands!
 			.filter(command => command.exec?.commandLine)
-			.filter(command => !command.attributes || (command.attributes as any)['controller.devfile.io/imported-by'] === undefined)
+			.filter(command => !command.attributes || ((command.attributes as any)['controller.devfile.io/imported-by'] === undefined || 'parent'))
```
l0rd:
```suggestion
			.filter(command => !command.attributes || (command.attributes as any)['controller.devfile.io/imported-by'] === (undefined || 'parent'))
```

RomanNikitenko (approved):
Commands from the parent devfile are available for running:

image

l0rd:
I checked on dogfooding and the problem is fixed 🎉

PR #211, opened 2023-04-17
fix: load webview static resources from local host instead of CDN

vitaliy-guliy:
### What does this PR do?

If configured, replaces the base URL of the webview static resources to be able to load them from the same host as VS Code instead of getting them from CDN.
It makes the webviews working when the browser is behind a proxy.

To turn the redirection on, it needs to edit the devfile and defile `WEBVIEW_LOCAL_RESOURCES` environment variable for `tools` container. Defining the variable into the editor devfile temporary does not work as the environment variable is not transferred into the DevWorkspace object (needs to be double checked)
```
      env:
        - name: WEBVIEW_LOCAL_RESOURCES
          value: 'true'
```

When it configured to get resources from CDN, the browser console looks like following
Screenshot from 2023-04-17 19-34-41

When `WEBVIEW_LOCAL_RESOURCES` is set, `index.html`, `fake.html` and `service-worker.js` were taken from the same domain as main VS Code frame

Screenshot from 2023-04-17 19-29-43

https://issues.redhat.com/browse/CRW-3885
https://issues.redhat.com/browse/CRW-4178

#### Webview resources are local-workspace based:

1. Create a workspace from the repository https://github.com/vitaliy-guliy/web-nodejs-sample/tree/test-webviews-local-resources ( it will use prebuilt [quay.io/vgulyy/che-code:webviews-behind-a-proxy-stable](https://quay.io/vgulyy/che-code:webviews-behind-a-proxy-stable) editor image )

Before 2 and 3 steps, open the browser console, navigate to Network.

2. Open README.md at the project root, click Open Preview at the right top. Check that `index.html`, `fake.html` and `service-worker.js` files were loaded from the current domain.

Pay attention on other files with URI like this
```
https://vscode-remote+devspaces-002eapps-002esandbox-002dm3-002e1530-002ep1-002eopenshiftapps-002ecom.vscode-resource.vscode-cdn.net/checode/checode-linux-libc/extensions/github/markdown.css
```
Despite they have `vscode-cdn.net` in the URI, they are loaded from the local domain. Browser do not send any request to `vscode-cdn.net` for that files because of service worker catches the requests for those files and forwards them to the local domain.

3. Install Jira and Bitbucket extension, open Atlassian Settings tab ( webview based ), check that it works.

#### Webview resources are CDN-based:

The same as above, but create a workspace from https://github.com/vitaliy-guliy/web-nodejs-sample/tree/test-webviews-resources-on-cdn
Check the browser console, `index.html`, `fake.html` and `service-worker.js` files must be loaded from [vscode-cdn.net](vscode-cdn.net)

### DIFF
--- a/.rebase/replace/code/src/vs/workbench/contrib/webview/browser/pre/index.html.json
+++ b/.rebase/replace/code/src/vs/workbench/contrib/webview/browser/pre/index.html.json
@@ -1,0 +1,10 @@
+[
+	{
+		"from": "content=\\\"default-src 'none'; script-src 'sha256-............................................' 'self'; frame-src 'self'; style-src 'unsafe-inline';\\\"",
+		"by": "content=\\\"default-src 'none'; script-src 'unsafe-inline' 'self'; frame-src 'self'; style-src 'unsafe-inline';\\\""
+	},
+	{
+		"from": "const hostname = location.hostname;",
+		"by": "const hostname = location.hostname;\\\n\\\n\\\t\\\t\\\t\\\t// Run if we are on the same host.\\\n\\\t\\\t\\\t\\\tconst parent = new URL(parentOrigin);\\\n\\\t\\\t\\\t\\\tif (parent.hostname === hostname) {\\\n\\\t\\\t\\\t\\\t\\\treturn start(parentOrigin);\\\n\\\t\\\t\\\t\\\t}"
+	}
+]
--- a/build/scripts/entrypoint-volume.sh
+++ b/build/scripts/entrypoint-volume.sh
@@ -66,6 +66,51 @@ if [ -n "${OPENVSX_REGISTRY_URL+x}" ]; then
   sed -i -e "s|serviceUrl:\".*\",itemUrl:\".*\"},version|serviceUrl:\"${OPENVSX_URL}/gallery\",itemUrl:\"${OPENVSX_URL}/item\"},version|" out/vs/workbench/workbench.web.main.js
 fi
 
+
+# To switch the webview content being loaded from local workspace we need to check for 'WEBVIEW_LOCAL_RESOURCES' environment variable.
+if [ -n "${WEBVIEW_LOCAL_RESOURCES+x}" ]; then
+
+  # If the variable is set to 'true', then:
+  if [[ "${WEBVIEW_LOCAL_RESOURCES}" == "true" ]]; then
+
+    # fetch che-code endpoint value from /devworkspace-metadata/flattened.devworkspace.yaml
+    CHE_CODE_ENDPOINT=$(yq -r '
+      .components
+      | to_entries[]
+      | select(.value.container.command[0] == "/checode/entrypoint-volume.sh")
+      | .value.container.endpoints
+      | to_entries[]
+      | select(.value.name == "che-code")
+      | .value.attributes
+      | to_entries[]
+      | select(.key == "controller.devfile.io/endpoint-url")
+      | .value
+    ' /devworkspace-metadata/flattened.devworkspace.yaml)
+
+    # make a valid base URI for webview static resources
+    WEBVIEW_CONTENT_BASE_URL=$CHE_CODE_ENDPOINT"oss-dev/static/out/vs/workbench/contrib/webview/browser/pre/"
+    echo "WEBVIEW_CONTENT_BASE_URL: $WEBVIEW_CONTENT_BASE_URL"
+
+    # fetch the existing base URI from product.json
+    WEBVIEW_CONTENT_BASE_URL_TEMPLATE=$(jq -r '.webviewContentExternalBaseUrlTemplate' product.json)
+    echo "WEBVIEW_CONTENT_BASE_URL_TEMPLATE: $WEBVIEW_CONTENT_BASE_URL_TEMPLATE"
+
+    # escape special characters
+    WORK_TEMPLATE=$WEBVIEW_CONTENT_BASE_URL_TEMPLATE
+    WORK_TEMPLATE=${WORK_TEMPLATE//[\{]/\\{}
+    WORK_TEMPLATE=${WORK_TEMPLATE//[\}]/\\\}}
+
+    # apply new base URI to the following files
+    sed -i -r -e "s|${WORK_TEMPLATE}|${WEBVIEW_CONTENT_BASE_URL}|" out/vs/workbench/workbench.web.main.js
+    sed -i -r -e "s|${WORK_TEMPLATE}|${WEBVIEW_CONTENT_BASE_URL}|" out/vs/workbench/api/node/extensionHostProcess.js
+
+    # actual value of webviewContentExternalBaseUrlTemplate in the product.json has no effect at runtime
+    # it serves as a marker for easily verifying the source of the webview static resources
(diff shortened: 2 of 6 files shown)


code/src/vs/workbench/contrib/webview/browser/pre/index-no-csp.html
```diff
@@ -333,6 +333,12 @@
 
 				const hostname = location.hostname;
 
+				// Run if we are on the same host.
+				const parent = new URL(parentOrigin);
+				if (parent.hostname === hostname) {
+					return start(parentOrigin);
+				}
+
```
RomanNikitenko, line 341: I think it makes sense to provide a logic for automatic rebase
@azatsarynnyy @vitaliy-guliy wdyt?

code/src/vs/workbench/contrib/webview/browser/pre/index.html
```diff
@@ -334,6 +334,12 @@
 
 				const hostname = location.hostname;
 
+				// Run if we are on the same host.
+				const parent = new URL(parentOrigin);
+				if (parent.hostname === hostname) {
+					return start(parentOrigin);
+				}
+
```
RomanNikitenko, line 342: please see https://github.com/che-incubator/che-code/pull/211/files#r1169814221

build/scripts/entrypoint-volume.sh
```diff
@@ -66,6 +66,44 @@ if [ -n "${OPENVSX_REGISTRY_URL+x}" ]; then
   sed -i -e "s|serviceUrl:\".*\",itemUrl:\".*\"},version|serviceUrl:\"${OPENVSX_URL}/gallery\",itemUrl:\"${OPENVSX_URL}/item\"},version|" out/vs/workbench/workbench.web.main.js
 fi
 
+
+# To switch the webview content being loaded from local workspace we need to check for 'WEBVIEW_LOCAL_RESOURCES' environment variable.
+# If the variable is set, then:
+if [ -n "${WEBVIEW_LOCAL_RESOURCES+x}" ]; then
```
azatsarynnyy, line 71: In this ^^ case, the actual value doesn't matter here. Only the environment variable presence is checked. Correct?
Wouldn't it be clearer if we checked for a specific value (`'true'`)?
In all other cases - the default behavior.

build/scripts/entrypoint-volume.sh
```diff
@@ -66,6 +66,44 @@ if [ -n "${OPENVSX_REGISTRY_URL+x}" ]; then
...
+
+  # escape special characters
+  WORK_TEMPLATE=$WEBVIEW_CONTENT_BASE_URL_TEMPLATE
+  WORK_TEMPLATE=${WORK_TEMPLATE//[\{]/\\{}
+  WORK_TEMPLATE=${WORK_TEMPLATE//[\}]/\\\}}
+
+  # apply new base URI to the following files
+  sed -i -r -e "s|${WORK_TEMPLATE}|${WEBVIEW_CONTENT_BASE_URL}|" out/vs/workbench/workbench.web.main.js
+  sed -i -r -e "s|${WORK_TEMPLATE}|${WEBVIEW_CONTENT_BASE_URL}|" out/vs/workbench/api/node/extensionHostProcess.js
+  sed -i -r -e "s|${WORK_TEMPLATE}|${WEBVIEW_CONTENT_BASE_URL}|" product.json
```
azatsarynnyy: To make it clear, I'm suggesting adding a short comment here that the actual value of `webviewContentExternalBaseUrlTemplate` in the `product.json` has no effect at runtime, and it serves as a marker for easily verifying the source of the webview static resources.

code/src/vs/workbench/contrib/webview/browser/pre/index-no-csp.html
azatsarynnyy: Considering that [it's changing about every 2 weeks](https://github.com/microsoft/vscode/commits/main/src/vs/workbench/contrib/webview/browser/pre/index.html), it would be nice [to add a rebase rule](https://github.com/che-incubator/che-code/blob/main/rebase.sh#L219) for resolving the potential merge conflicts automatically, so fewer distractions for us.

build/scripts/entrypoint-volume.sh
vitaliy-guliy: Yes, you are right. I thought about checking for both `true` or `yes` here.
User could set `no` which would be confusing.

azatsarynnyy: `true/false` are more common values.
E.g.: https://github.com/eclipse-che/che-plugin-registry/blob/7c0ad3e1f72e7ea521156cfb98ed80c5d657991b/che-editors.yaml#L57-L66

.rebase/replace/code/src/vs/workbench/contrib/webview/browser/pre/index.html.json
```diff
@@ -0,0 +1,6 @@
+[
+	{
+		"from": "<meta http-equiv=\"Content-Security-Policy\"\\\n",
+		"by": "--->"
+	}
```
RomanNikitenko, line 9: the corresponding handler should be added to the `rebase.sh` script for the `code/src/vs/workbench/contrib/webview/browser/pre/index.html` file to get it working

vitaliy-guliy: Done

code/src/vs/workbench/contrib/webview/browser/pre/index-no-csp.html
vitaliy-guliy: Done

code/src/vs/workbench/contrib/webview/browser/pre/index.html
vitaliy-guliy: Done

build/scripts/entrypoint-volume.sh
vitaliy-guliy: Done

build/scripts/entrypoint-volume.sh
vitaliy-guliy: Done

azatsarynnyy (approved):
I tested the PR for both cases:
1/ the default one, when the webview resources are loaded from the CDN (using the [image built for this PR](quay.io/che-incubator-pull-requests/che-code:pr-211-amd64) by the GH action);
2/ when `WEBVIEW_LOCAL_RESOURCES` is set and webview-related files are loaded from the Che-Code origin URL (using the [image provided](https://quay.io/vgulyy/che-code:webviews-behind-a-proxy-stable) in the PR's testing steps).

I can confirm that both options work as expected :+1: 

In the 2nd case, I was able to open the markdown preview which is webview-based. I believe it's enough to make such a simple test to verify that this patch solves both problems [CRW-3885](https://issues.redhat.com/browse/CRW-3885) and [CRW-4178](https://issues.redhat.com/browse/CRW-4178).

##### How did I test it?
To make sure the 2nd option works as expected when the user's browser is in a proxied environment, I configured the system proxy settings on my local Fedora as follows:

image

There's no real proxy server on `localhost`. So, all the outbound connections are blocked.
Except for the Che host itself which I added to the `Ignore Hosts`.

PR #214, opened 2023-04-19
feat: Support Devfile exec commands env in VS Code tasks

RomanNikitenko:
### What does this PR do?
- Support Devfile exec commands env in VS Code tasks:
         resolve env variables for exec commands at `devfile command => VS Code task` transformation step
- Turn On execution `jest` tests for the `che-commands` extension 
 

https://github.com/eclipse/che/issues/21760

### How to test this PR?
- you can use [this link](https://che-dogfooding.apps.che-dev.x6e0.p1.openshiftapps.com/#https://github.com/RomanNikitenko/web-nodejs-sample/tree/test-env-var-for-commands) to start a workspace
- [it's configured](https://github.com/RomanNikitenko/web-nodejs-sample/blob/5a20412ce02c0b17749db5a48c598f78875cca1f/.che/che-editor.yaml#L33) to use image of current PR
- there are [few commands](https://github.com/RomanNikitenko/web-nodejs-sample/blob/5a20412ce02c0b17749db5a48c598f78875cca1f/devfile.yaml#L26-L61) in the devfile for testing
- go to the `Terminal` => `Run Task` `devfile` and select a task for testing

### DIFF
--- a/code/extensions/che-commands/src/taskProvider.ts
+++ b/code/extensions/che-commands/src/taskProvider.ts
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (c) 2022 Red Hat, Inc.
+ * Copyright (c) 2022-2023 Red Hat, Inc.
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
@@ -10,8 +10,9 @@
 
 /* eslint-disable header/header */
 
-import { V1alpha2DevWorkspaceSpecTemplate, V1alpha2DevWorkspaceSpecTemplateCommands } from '@devfile/api';
+import { V1alpha2DevWorkspaceSpecTemplate, V1alpha2DevWorkspaceSpecTemplateCommands, V1alpha2DevWorkspaceSpecTemplateCommandsItemsExecEnv } from '@devfile/api';
 import * as vscode from 'vscode';
+import { resolveEnvVariablesForCommand } from './utils';
 
 interface DevfileTaskDefinition extends vscode.TaskDefinition {
 	command: string;
@@ -38,7 +39,7 @@ export class DevfileTaskProvider implements vscode.TaskProvider {
 		const cheTasks: vscode.Task[] = devfileCommands!
 			.filter(command => command.exec?.commandLine)
 			.filter(command => !command.attributes || (command.attributes as any)['controller.devfile.io/imported-by'] === undefined)
-			.map(command => this.createCheTask(command.exec?.label || command.id, command.exec?.commandLine!, command.exec?.workingDir || '${PROJECT_SOURCE}', command.exec?.component!));
+			.map(command => this.createCheTask(command.exec?.label || command.id, command.exec?.commandLine!, command.exec?.workingDir || '${PROJECT_SOURCE}', command.exec?.component!, command.exec?.env));
 		return cheTasks;
 	}
 
@@ -52,7 +53,7 @@ export class DevfileTaskProvider implements vscode.TaskProvider {
 		return [];
 	}
 
-	private createCheTask(name: string, command: string, workdir: string, component: string): vscode.Task {
+	private createCheTask(name: string, command: string, workdir: string, component: string, env?: Array<V1alpha2DevWorkspaceSpecTemplateCommandsItemsExecEnv>): vscode.Task {
 		function expandEnvVariables(line: string): string {
 			const regex = /\${[a-zA-Z_][a-zA-Z0-9_]*}/g;
 			const envArray = line.match(regex);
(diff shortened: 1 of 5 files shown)


code/extensions/che-commands/src/taskProvider.ts
```diff
@@ -75,9 +75,38 @@ export class DevfileTaskProvider implements vscode.TaskProvider {
 		};
 
 		const execution = new vscode.CustomExecution(async (): Promise<vscode.Pseudoterminal> => {
-			return this.terminalExtAPI.getMachineExecPTY(component, command, expandEnvVariables(workdir));
+			return this.terminalExtAPI.getMachineExecPTY(component, resolveEnvVariablesForCommand(command, env), expandEnvVariables(workdir));
```
RomanNikitenko, line 79: @azatsarynnyy 
I would like to know your opinion here:
- is it enough to resolve such variables on this step?
so, we resolve `command` when `TaskProvider` creates tasks, but we don't set such env variables to the corresponding container 
 or
- the issue assumes that such variables should be set to the corresponding container and available from a terminal as well?

code/extensions/che-commands/src/taskProvider.ts
```diff
@@ -75,9 +75,38 @@ export class DevfileTaskProvider implements vscode.TaskProvider {
...
 		});
 		const task = new vscode.Task(kind, vscode.TaskScope.Workspace, name, 'devfile', execution, []);
 		return task;
 	}
 }
+
+function resolveEnvVariablesForCommand(commandLine: string, envVariables?: Array<V1alpha2DevWorkspaceSpecTemplateCommandsItemsExecEnv>): string {
```
RomanNikitenko: I'll try to simplify this function...

code/extensions/che-commands/src/taskProvider.ts
azatsarynnyy: [Devfile spec v. 2.2.0](https://devfile.io/docs/2.2.0/devfile-schema#commands-exec-env) says that a command's env. variables list is
> `Optional list of environment variables that have to be set before running the command`

I understand it as such variables should be set only for a specific command execution.

azatsarynnyy: > the issue assumes that such variables should be set to the corresponding container and available from a terminal as well?

it's supported by a container component's own `env` property 
https://devfile.io/docs/2.2.0/devfile-schema#components-container-env

code/extensions/che-commands/src/taskProvider.ts
RomanNikitenko: I've applied some changes to simplify it

RomanNikitenko:
The functionality is based on the regular expression usage, so I'm working on adding some tests, will update my PR soon.

code/extensions/che-commands/src/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ ***********************************************************************/
+
+/* eslint-disable header/header */
+
+import { V1alpha2DevWorkspaceSpecTemplateCommandsItemsExecEnv } from "@devfile/api";
+
+/*
+ * This function is used to resolve env variables in command line.
+ * It should support both syntaxes: ${VAR} and $VAR
```
azatsarynnyy, line 17: It should support or it really supports? )

code/extensions/che-commands/src/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+	if (!envVariables || envVariables.length === 0) {
+		return commandLine;
+	}
+
+	// it should support both syntaxes: ${VAR} and $VAR
+	const regex = /\$\{?([a-zA-Z_][a-zA-Z0-9_]*)\}?/g;;
+	let match = regex.exec(commandLine);
+	while (match) {
+		const envVariableName = match[1];
+		const envVariableValue = envVariables.find(env => env.name === envVariableName)?.value;
```
azatsarynnyy, line 33: `envVariables` array contains only the variables defined in the devfile. It doesn't read from `process.env`, correct?
I didn't get how resolving the "real" environment variables works (like `KUBERNETES_SERVICE_HOST`).
Could you point me out, please?

code/extensions/che-commands/src/utils.ts
RomanNikitenko: the answer is in tests 😉

code/extensions/che-commands/src/utils.ts
RomanNikitenko: let's say a command looks like:
`echo "KUBERNETES_SERVICE_HOST ${KUBERNETES_SERVICE_HOST}"`

for such use case function `resolveEnvVariablesForCommand()` should return the command as is:
`echo "KUBERNETES_SERVICE_HOST ${KUBERNETES_SERVICE_HOST}"`

the env variable is resolved directly in the terminal if it's defined.
try to execute the same command in a terminal for the `tools` container: 
image

so, there is no magic  :-)

azatsarynnyy: Got it. Thanks!

azatsarynnyy (approved):
Thanks @RomanNikitenko for providing the clear testing steps.
Everything works as expected :+1:

PR #219, opened 2023-04-28
fix: Fix pattern for replacing OPENVSX_URL variable

RomanNikitenko:
### What does this PR do?
Fixes pattern for replacing `OPENVSX_URL` variable in the `/checode/checode-linux-libc/out/vs/workbench/workbench.web.main.js` file.
The pattern is used in the entrypoint, see [here](https://github.com/che-incubator/che-code/blob/4cc3fd0d4338e52ed88cdb8364b93fe7ccd32392/build/scripts/entrypoint-volume.sh#L65).

Before this change the pattern broke the corresponding product configuration - it [cut all values of the product.json file](https://github.com/redhat-developer/devspaces-images/blob/9b48e7a56904dfd60de0c96b32b3131511cfc9b9/devspaces-code/code/product.json#L101-L125) which were placed after the `serviceUrl` and `itemUrl` fields. 

https://github.com/eclipse/che/issues/21962
https://issues.redhat.com/browse/CRW-4114

### How to test this PR?
I tested it in the downstream:
- start a workspace
- open `/checode/checode-linux-libc/out/vs/workbench/workbench.web.main.js` file
- search `extensionsGallery:`
- check that [all fields of the product.json file](https://github.com/redhat-developer/devspaces-images/blob/9b48e7a56904dfd60de0c96b32b3131511cfc9b9/devspaces-code/code/product.json#L101-L125) are present in the file

Screenshot 2023-05-05 at 20 08 54

The pattern should also handle correctly extra fields in the `extensionGalary`, for example:
```
"extensionsGallery": {
    "serviceUrl": "https://open-vsx.org/vscode/gallery",
    "itemUrl": "https://open-vsx.org/vscode/item",
    "resourceUrlTemplate": "someResourceUrlTemplate"
  },
```

### DIFF
--- a/build/scripts/entrypoint-volume.sh
+++ b/build/scripts/entrypoint-volume.sh
@@ -62,7 +62,7 @@ if [ -n "${OPENVSX_REGISTRY_URL+x}" ]; then
   echo "using OPENVSX_URL=$OPENVSX_URL"
   sed -i -r -e "s|\"serviceUrl\": \"..*\"|\"serviceUrl\": \"${OPENVSX_URL}/gallery\"|" product.json
   sed -i -r -e "s|\"itemUrl\": \"..*\"|\"itemUrl\": \"${OPENVSX_URL}/item\"|" product.json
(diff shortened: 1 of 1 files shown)


azatsarynnyy (approved):
Nice catch :+1:

PR #221, opened 2023-05-03
chore: Remove delay at starting Machine-Exec step

RomanNikitenko:
### What does this PR do?
Remove a delay at starting Machine-Exec step, it should reduce starting VS Code process by 5 seconds.

https://issues.redhat.com/browse/CRW-4342

### How to test this PR?
- create a terminal in a container directly after starting VS Code 
- check that the terminal works as expected

### DIFF
--- a/build/scripts/entrypoint-volume.sh
+++ b/build/scripts/entrypoint-volume.sh
@@ -25,7 +25,6 @@ ls -la /checode/
 # Start the machine-exec component in background
 export MACHINE_EXEC_PORT=3333
 nohup /checode/bin/machine-exec --url "0.0.0.0:${MACHINE_EXEC_PORT}" &
-sleep 5
 
 # Start the checode component based on musl or libc
 


build/scripts/entrypoint-volume.sh
```diff
@@ -25,7 +25,6 @@ ls -la /checode/
 # Start the machine-exec component in background
 export MACHINE_EXEC_PORT=3333
 nohup /checode/bin/machine-exec --url "0.0.0.0:${MACHINE_EXEC_PORT}" &
-sleep 5
```
RomanNikitenko, line 28: @azatsarynnyy 
I've found this delay within investigation of another issue. I'm not sure why do we need it and if we can remove it

azatsarynnyy: @RomanNikitenko I agree, it looks redundant now.
I suppose the history of this `sleep` is something like this:

It was introduced within the initial implementation. At that time, Che-Code started requesting machine-exec during the Code server launch. Probably, some artificial delay was required because machin-exec wasn't ready to serve the clients.
Since I reworked communication with machine-exec and it's requested only after the corresponding built-in extension (che-terminal) has started, I believe we don't need that timeout anymore. As machine-exec has enough time to start now.

azatsarynnyy (approved):

build/scripts/entrypoint-volume.sh
```diff
@@ -25,7 +25,6 @@ ls -la /checode/
 # Start the machine-exec component in background
 export MACHINE_EXEC_PORT=3333
 nohup /checode/bin/machine-exec --url "0.0.0.0:${MACHINE_EXEC_PORT}" &
```
azatsarynnyy, line 27: @RomanNikitenko it would be nice to remove the trailing `&` which is redundant now.

RomanNikitenko: > If a command is terminated by the control operator &, the shell executes the command in the background in a subshell. The shell does not wait for the command to finish, and the return status is 0.


do we really want to wait starting `machine-exec` here?
do you see a real use case when a terminal is not ready after starting IDE?

azatsarynnyy: you're right - running it asynchronously sounds perfect
Please, ignore my note. Sorry for the noise.

PR #227, opened 2023-05-15
Don't run the tests before the dependencies are actually installed

azatsarynnyy:
### What does this PR do?
Recently, running the test for the `che-commands` extensions [was bound](https://github.com/che-incubator/che-code/commit/319c20aa848e32c2c5e49b1ccbac2acd1b4c128d) before the dependencies installation phase.
While it works well when building Che-Code assembly from the root of the project, running the dependencies installation solely for the `che-commands` extension fails. It's because the script depends on the `gulp` binary which isn't installed yet.
It affects the downstream build.

After merging this PR, the tests for `che-commands` extensions can be run the same way as for other extensions - by calling `yarn test` command, e.g: https://github.com/che-incubator/che-code/blob/8cf7bbc8a81ac46e6162b81900906e89babe1480/code/extensions/html-language-features/server/package.json#L30

It's needed for https://issues.redhat.com/browse/CRW-3160

### How to test this PR?
After running `cd code/extensions/che-commands && yarn`
the extension's dependencies should be installed successfully, without the error:
```
$ gulp compile-extension:che-commands
/bin/sh: line 1: gulp: command not found
error Command failed with exit code 127.
```

### DIFF
--- a/code/extensions/che-commands/package.json
+++ b/code/extensions/che-commands/package.json
@@ -23,7 +23,6 @@
   },
   "main": "./out/extension.js",
   "scripts": {
-    "prepare": "yarn compile && yarn test",
     "compile": "gulp compile-extension:che-commands",
     "watch": "gulp watch-extension:che-commands",
     "vscode:prepublish": "npm run compile",


code/extensions/che-commands/package.json
```diff
@@ -23,7 +23,6 @@
   },
   "main": "./out/extension.js",
   "scripts": {
-    "prepare": "yarn compile && yarn test",
```
RomanNikitenko, line 26: so, the solution is - just turn off tests at all?

RomanNikitenko:
> After merging this PR, the tests for che-commands extensions can be run the same way as for other extensions - by calling yarn test command

just for my understanding - when do tests for other extensions run?
- on the project build step?
- on the docker build step?
- on the github actions step?
- some another step?

FYI: some of them are already broken, so I guess they are never run:

image

code/extensions/che-commands/package.json
azatsarynnyy: The short answer: to fix the failed downstream build we need to make the dependencies installation work again - as it used to be a few weeks ago.

@RomanNikitenko I see your concern. Let me explain the problem from a bit different perspective with more details.
We have in-progress work on the new downstream build approach - with Cachito. It requires running the dependencies installation (calling `yarn install`) for each VS Code sub-module separately. Once the dependencies are installed for each sub-module, we can run the `gulp` build as the `gulp` binary already installed. It used to work as expected until we introduced [this change](https://github.com/che-incubator/che-code/commit/319c20aa848e32c2c5e49b1ccbac2acd1b4c128d) recently. The mentioned change means that there's no more a way to just install the dependencies solely for the `che-commands` sub-package. Moreover, unlike other sub-packages, calling `yarn install` for `che-commands` now triggers all three steps:
- dependencies installation;
- actual build;
- running tests.

In other words, when one tries just to install the dependencies for the `che-commands` sub-module, `yarn install` command always fails.

So, this PR just reverts the regression introduced by the recent change to fix the failed downstream build.
Once we migrate to the new downstream build approach we will be able to detect such problems much faster. As our downstream build job will comment on the corresponding Che-Code PR immediately that the downstream build has failed.

To separate the broken extension's dependencies installation problem from the problem of automating running the tests (instead of calling `yarn test` manually) we can create a separate issue to discuss the ideas for the second problem. Will it work well for you, @RomanNikitenko?

RomanNikitenko: > to fix the failed downstream build we need to make the dependencies installation work again - as it used to be a few weeks ago.

The corresponding jobs are green for the downstream build:
image


So, I guess, it's more about:
> We have in-progress work on the new downstream build approach - with Cachito. It requires running the dependencies installation (calling yarn install) for each VS Code sub-module separately.

Correct?

I understand that running tests should be moved on the another stage, but, I do believe we should not just turn off them. It's a very strange situation that **we have tests** for our extensions (btw, not only for the `che-commands`) and **we don't run them at all**: https://github.com/che-incubator/che-code/pull/227#issuecomment-1548387927.

About a separate issue to turn on tests for all our extensions - yes - it works for me, I hope the issue will get a high priority, otherwise it means that our tests are going to have quite long holiday :-)

RomanNikitenko (approved):

code/extensions/che-commands/package.json
azatsarynnyy: > > So, I guess, it's more about:
> 
> We have in-progress work on the new downstream build approach - with Cachito. It requires running the dependencies installation (calling yarn install) for each VS Code sub-module separately.
> 
> Correct?

Correct.

azatsarynnyy:
> just for my understanding - when do tests for other extensions run?

I suppose the Che-specific extensions were just ported from Che-Theia as it is.  I don't remember someone working on any GitHub Action for running the tests. The only way I know is to run `yarn test` command manually. That's why my suggestion is to automate that with a GitHub Action.
Thanks for the filed issue https://github.com/eclipse/che/issues/22213 @RomanNikitenko!

RomanNikitenko:
@azatsarynnyy 
btw, after merging your PR I still have a problem with the `gulp` dependency:
image

So, I'm able to provide a change to the extension, but I'm not able to compile only this extension - I have to build all assembly...

azatsarynnyy:
It fixes the issue of installing the extension's dependencies. In your example, running `yarn` command finishes successfully now :tada: 

I've also never seen the `compile` scripts (from package.json files) used directly in VS Code :man_shrugging: 
@RomanNikitenko  Usually, if I need to compile any single built-in extension of VS Code, I use [the `compile-extension:` gulp task they provide](https://github.com/microsoft/vscode/blob/46c9471450f82ef2500d86c555504502652d9aa5/build/gulpfile.extensions.js#L166). E.g.: `npx gulp compile-extension:che-commands` (from the VS Code root, of course):
```
$ npx gulp compile-extension:che-commands
[11:11:39] Using gulpfile ~/github/che-incubator/che-code/code/gulpfile.js
[11:11:39] Starting 'compile-extension:che-commands'...
[11:11:39] Starting clean-extension-che-commands ...
[11:11:39] Finished clean-extension-che-commands after 5 ms
[11:11:39] Starting compile-extension:che-commands ...
[11:11:39] Starting compilation extensions...
[11:11:41] Finished compilation extensions with 0 errors after 1475 ms
[11:11:41] Finished compile-extension:che-commands after 1737 ms
[11:11:41] Finished 'compile-extension:che-commands' after 1.75 s
```

PR #229, opened 2023-05-17
chore: Remove unused env variable

RomanNikitenko:
### What does this PR do?
Removes code related to the `CHE_PLUGIN_REGISTRY_INTERNAL_URL` env variable.
I don't see why we need it on the `che-code` side. 

no issue - just clean up

### How to test this PR?
Tests should be happy.

### DIFF
--- a/code/extensions/che-api/src/impl/k8s-devworkspace-env-variables.ts
+++ b/code/extensions/che-api/src/impl/k8s-devworkspace-env-variables.ts
@@ -55,13 +55,6 @@ export class K8sDevWorkspaceEnvVariables {
   private readonly pluginRegistryURL!: string;
 
   /**
-   * pluginRegistryInternalURL - Plugin registry internal URL
-   */
-
-  private readonly pluginRegistryInternalURL!: string;
-
-
-  /**
    * dashboardURL - Dashboard URL
    */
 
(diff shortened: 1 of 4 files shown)


code/extensions/che-api/src/impl/k8s-devworkspace-env-variables.ts
```diff
@@ -104,12 +97,6 @@ export class K8sDevWorkspaceEnvVariables {
       this.pluginRegistryURL = process.env.CHE_PLUGIN_REGISTRY_URL;
     }
 
-    if (process.env.CHE_PLUGIN_REGISTRY_INTERNAL_URL === undefined) {
```
RomanNikitenko, line 107: @svor @azatsarynnyy @vitaliy-guliy  
please let me know if we still should support the variable

svor: I don't see why we need this env in che-code

azatsarynnyy (approved):

code/extensions/che-api/src/impl/k8s-devworkspace-env-variables.ts
RomanNikitenko: thank you!

svor: giphy

vitaliy-guliy (approved):
If it is not used anymore, let's remove it

PR #240, opened 2023-06-12
fix: plugin endpoints should be visible by default

vitaliy-guliy:
### What does this PR do?
Makes plugins endpoints visible by default

Screenshot from 2023-06-12 16-38-01

https://issues.redhat.com/browse/CRW-4521

### How to test this PR?
Two ways:

The first, is to wait for che-code image, that will be build by this pull request and use it to test the changes

Second (preferable), use Che instance to
- create a workspsce from the branch
- run `prepare` and then `build` devfile commands
- then run `run` command to start server
- use the notification with proposal to open a new che-code instance in separate browser tab
- open to Endpoints view and check the endpoints

### DIFF
--- a/code/extensions/che-port/package.json
+++ b/code/extensions/che-port/package.json
@@ -63,7 +63,7 @@
     "views": {
       "endpoints": [
         {
-          "name": "endpoints",
+          "name": "Endpoints",
           "id": "endpoints"
         }
       ]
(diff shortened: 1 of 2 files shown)


RomanNikitenko:
Tested, it works for me:

image


https://github.com/che-incubator/che-code/assets/5676062/66444cb8-3aa1-41ae-ba8f-24c847f64e17

RomanNikitenko (approved):

PR #250, opened 2023-06-26
Provide Node.js for running VS Code

RomanNikitenko:
### What does this PR do?
Provides `Node.js` for running VS Code

https://github.com/eclipse/che/issues/21778


RomanNikitenko:
Another approach was applied within PR: https://github.com/che-incubator/che-code/pull/324

PR #252, opened 2023-06-29
Fix an automatic rebase against the upstream

### LINKED ISSUE #22296: Che-Code automatic rebase against upstream VS Code is failed

[The GitHub Workflow for Che-Code automatic rebase against upstream VS Code](https://github.com/che-incubator/che-code/actions/runs/5267193542/jobs/9522131297) failed with:
```
#13 1562.7 [13:10:10] Error: /checode-compilation/src/vs/workbench/contrib/extensions/browser/extensionRecommendationNotificationService.ts(273,12): Property 'installExtensions' does not exist on type '$H9'. Did you mean 'onInstallExtension'?
```

Fix the incompatibility with the upstream.

_No response_

### Additional context
The related upstream changes: https://github.com/microsoft/vscode/commit/f3aa9a201b2455e6556797e8ffb7145d77adb792 + https://github.com/microsoft/vscode/commit/debcf16fcd8896f598aada07c19b4cc461bdbd18

azatsarynnyy:
### What does this PR do?
Fixes an auto-rebase against the upstream.

closes https://github.com/eclipse/che/issues/22296

### How to test this PR?
`Pull Request Check / build ...` PR status checks must be green.


code/extensions/mangle-loader.js
```diff
@@ -25,7 +25,7 @@ function getMangledFileContents(projectPath) {
 	if (!entry) {
 		const log = (...data) => fancyLog(ansiColors.blue('[mangler]'), ...data);
 		log(`Mangling ${projectPath}`);
-		const ts2tsMangler = new Mangler(projectPath, log, { mangleExports: true, manglePrivateFields: true });
+		const ts2tsMangler = new Mangler(projectPath, log, { mangleExports: false, manglePrivateFields: true });
 		entry = ts2tsMangler.computeNewFileContents();
```
RomanNikitenko, line 29: Sorry, could you explain why we should change default VS Code behaviour here?

rebase.sh
```diff
@@ -300,6 +313,8 @@ resolve_conflicts() {
       apply_code_package_changes
     elif [[ "$conflictingFile" == "code/product.json" ]]; then
       apply_code_product_changes
+    elif [[ "$conflictingFile" == "code/extensions/mangle-loader.js" ]]; then
```
RomanNikitenko: does the PR contain the corresponding part in the `.rebase` folder?
this PR contains a lot of commits, so maybe I just missed it...

RomanNikitenko:
I guess that the current changes are related to the `Mangler` logic.
image 

If so - it's related to reducing code loading times. As far as I understand it's critical for our product ([CRW-4342](https://issues.redhat.com//browse/CRW-4342)).
So, it's better to keep upstream changes, in few weeks no one will remember that this part is different from the VS Code and can be improved... 

In general - maybe we should document every patch of the VS Code with the reference to the corresponding commit and short description - **why** we had to change the original logic.

rebase.sh
azatsarynnyy: > maybe I just missed it...

Good catch! :+1: Thanks a lot for noticing the missing file I forgot to commit :facepalm: 
Added a new replacement rule: https://github.com/che-incubator/che-code/pull/252/commits/396fba65f4f850721017d32a8f85a5c7f64ec85f

azatsarynnyy:
> ...it's related to reducing code loading times. As far as I understand it's critical for our product...

Exactly. For about half a year Che-Code already benefits from those [upstream patches](https://github.com/microsoft/vscode/pull/166126/files) that were part of the [November's release](https://code.visualstudio.com/updates/v1_74#_engineering). 

> ...in few weeks no one will remember that this part is different from the VS Code and can be improved...

Thanks @RomanNikitenko! It's a very good point. I filed https://github.com/eclipse/che/issues/22337 to address this item.

> In general - maybe we should document every patch of the VS Code with the reference to the corresponding commit and short description - why we had to change the original logic.

Good idea :+1: Feel free to describe it in an issue. Maybe, someone will take it for a H'n'H Day.

code/extensions/mangle-loader.js
azatsarynnyy: Because the mangler fails for che-api extension: https://github.com/eclipse/che/issues/22296#issuecomment-1597383783
Even when the exports are annotated with `@skipMangle`.
If we can make time for it, I filed [the issue](https://github.com/eclipse/che/issues/22337) to dig into the implementation details of VS Code mangling process and investigate it.

RomanNikitenko:
> Exactly. For about half a year Che-Code already benefits from those [upstream patches](https://github.com/microsoft/vscode/pull/166126/files) that were part of the [November's release](https://code.visualstudio.com/updates/v1_74#_engineering).

yes, that one was for the November's release
but now they do the next step and apply similar approach for the extensions
or 
do you think that it's less important part?

my point of view is - we should not ignore `Mangler` related logic, we should not just turn off it...

azatsarynnyy:
> do you think that it's less important part?

@RomanNikitenko sure, it's less important. The main goal is to unblock consuming the upstream patches ASAP and continue moving forward with our other priorities.
That "nice to have" enhancement (`mangleExports`) can be enabled at any time in the future (https://github.com/eclipse/che/issues/22337) once we decide it's one of the highest priorities for the product.

RomanNikitenko:
> That "nice to have" enhancement (mangleExports) can be enabled at any time in the future (https://github.com/eclipse/che/issues/22337) once we decide it's one of the highest priorities for the product

Agree, but there was no issue, no comment, nothing in the PR description about it when I was reviewing it.
That's why I raised this question.
As I mentioned above:
> in few weeks no one will remember that this part is different from the VS Code and can be improved...

 Now we have an issue, thanks!

RomanNikitenko:
image

I don't think that it's something that we should do within H'n'H Day.
From my point of view we should just start to do it when we provide changes that differ from the upstream. 
it's easier to write a few lines that describe a patch when we provide it 
than 
search them in 1 year and think "Why do we have this difference?"

RomanNikitenko:
About https://github.com/eclipse/che/issues/22296#issuecomment-1597383783
We have the problem for 4 files in 1 che-extension.
But there are about 100 extensions...
Maybe we could say `Mangler` to ignore that 4 entities (4 interfaces)?

I mean it's better to ignore 4 entities in 1 extension instead of turning off `Mangler` logic for about 100 extensions...

RomanNikitenko:
> I mean it's better to ignore 4 entities in 1 extension instead of turning off Mangler logic for about 100 extensions...

btw,
I was able to achieve it, please see my idea [here](https://github.com/che-incubator/che-code/pull/255/commits/11c52c520ebd623c33c1448ab503c5092e5ae217)

As you can see 
- the Mangler is turned On for all extensions except 4 entities in the `che-api` extension  
- the jobs are happy

image


I can push my fix to your branch if you don't mind.

azatsarynnyy:
> Maybe we could say Mangler to ignore that 4 entities (4 interfaces)?

it's a good alternative with its own pros/cons. At least, it will narrow down the scope of exclusions

> I can push my fix to your branch if you don't mind.

I don't mind. Please do. Thanks @RomanNikitenko for looking at it

RomanNikitenko:
> I don't mind. Please do. Thanks @RomanNikitenko for looking at it

Done!
Please make sure that you have updated `rebase` rules according to my logic.

RomanNikitenko:
> it's a good alternative with its own pros/cons

just for my understanding - could you clarify what `cons` you see with my approach (if we compare to the previous one from the current PR)?

RomanNikitenko (approved):

RomanNikitenko:
@azatsarynnyy 
btw, your logic related to the automatic git-rebase for the mangler logic works incorrectly: https://github.com/eclipse/che/issues/22428

.rebase/replace/code/build/lib/mangle/index.js
```diff
@@ -0,0 +1,6 @@
+[
```
RomanNikitenko, line 1: I guess file has incorrect name: https://github.com/eclipse/che/issues/22428

.rebase/replace/code/build/lib/mangle/index.ts
```diff
@@ -0,0 +1,6 @@
+[
```
RomanNikitenko, line 1: the same: https://github.com/che-incubator/che-code/pull/252/files#r1287962785

.rebase/replace/code/build/lib/mangle/index.js
azatsarynnyy: right - renamed in https://github.com/che-incubator/che-code/pull/266

PR #246, opened 2023-06-22
Use Universal Developer Image as dev container

vitaliy-guliy:
### What does this PR do?
- Inherits che-code-dev image from Universal Developer Image
- Adds few commands to build che-code image with podman

### What issues does this PR fix?

https://github.com/eclipse/che/issues/22292

### How to test this PR?
Do following:
- use the dogfooding instance to create a workspace from the current branch https://github.com/che-incubator/che-code/tree/switch-to-udi
- use devfile commands to compile and run VS Code with `yarn`
- use devfile commands to compile and then build `che-code` image with `podman`

Should have like following:

Screenshot from 2023-06-26 17-50-43

Screenshot from 2023-06-26 17-49-35

---

### Note that:

- all the 30 commits will be squashed when merging. I haven't squashed them here because of 1) all the commits to the branch were performed using online che-code, as for me the squashing it not clear UX flow; 2) do not want to add an extra commit as it  triggers the github PR checks

- at the moment, `quay.io/che-incubator-pull-requests/che-code-dev:pr-246-dev-amd64` is used as a dev image. I plan to update it after merging of this pull request and when the corresponding che-dev image will be built by the appropriate github action.

PR #21, opened 2022-01-27
chore: introduce zsh

benoitf:
For development mode, brings zsh shell

it includes command completion and git support (and also kubernetes current context being displayed)

PR #253, opened 2023-06-30
Use universal developer image for dev container

vitaliy-guliy:
### What does this PR do?
- Inherits che-code-dev image from Universal Developer Image
- Adds few commands to build che-code image with podman

https://github.com/eclipse/che/issues/22292

### How to test this PR?
Do following:
- use the dogfooding instance to create a workspace from the current branch https://github.com/che-incubator/che-code/tree/switch-to-udi-merge-test
- use devfile commands to compile and run VS Code with `yarn`
- use devfile commands to compile and then build `che-code` image with `podman`

Screenshot from 2023-09-07 16-01-35

Should have like following:

Screenshot from 2023-06-26 17-50-43

Screenshot from 2023-06-26 17-49-35

---

### Note that:

Depends on https://github.com/devfile/developer-images/pull/117


vitaliy-guliy:
@azatsarynnyy @RomanNikitenko could you please guys review this pull request and approve or deny it?
It has been opened already two weeks, it would be nice if we move forward with updating of che-code dev image.

build/dockerfiles/dev.Dockerfile
```diff
@@ -6,81 +6,49 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 
-# https://quay.io/eclipse/che-machine-exec#^7\.
-FROM quay.io/eclipse/che-machine-exec:7.56.0 as machine-exec
+FROM quay.io/devfile/universal-developer-image:latest
```
RomanNikitenko, line 9: @vitaliy-guliy 
As I mentioned in your previous PR https://github.com/che-incubator/che-code/pull/246#discussion_r1245414076
I don't think we should bring UDI for the `Che-Code` development

But in general it's OK for me to merge the current PR, if the changes look good for other reviewers

azatsarynnyy:
@vitaliy-guliy am I understand it right that for fixing https://github.com/eclipse/che/issues/22292 it's enough to add the Devfile commands from this PR: `podman-compile-libc` and `podman-assemble-che-code`?
I tried to build an image with podman in the workspace started from che-code's `main` branch and it finished successfully:
image


_update:_ sorry, I used a wrong image

build/dockerfiles/dev.Dockerfile
```diff
@@ -6,81 +6,49 @@
...
 
-# https://registry.access.redhat.com/ubi8/ubi
-FROM registry.access.redhat.com/ubi8/ubi:8.6-990 AS ubi-micro-build
-RUN mkdir -p /mnt/rootfs
-RUN ARCH=$(uname -m) && yum install --installroot /mnt/rootfs libsecret curl make cmake gcc gcc-c++ python3 git git-core-doc openssh less wget \
-                 "https://rpmfind.net/linux/centos/8-stream/BaseOS/x86_64/os/Packages/libsecret-devel-0.18.6-1.el8.x86_64.rpm" \
-                 "https://rpmfind.net/linux/centos/8-stream/AppStream/x86_64/os/Packages/libxkbfile-1.1.0-1.el8.x86_64.rpm" \
-                 "https://rpmfind.net/linux/centos/8-stream/PowerTools/x86_64/os/Packages/libxkbfile-devel-1.1.0-1.el8.x86_64.rpm" \
-                 procps vi nano \
-                 "https://rpmfind.net/linux/centos/8-stream/BaseOS/${ARCH}/os/Packages/zsh-5.5.1-6.el8_1.2.x86_64.rpm" \
```
azatsarynnyy, line 20: Why do we need to remove zsh, kubectl from the Dev image?
It was added intentionally (by https://github.com/che-incubator/che-code/pull/21), for dev purposes

azatsarynnyy: okay, kubectl is pre-installed in UDI

vitaliy-guliy: Ok, it's not a problem to get it back.
But I want to say you by a secret, `$HOME/.zshrc` is never called.
That was probably a reason why I removed zsh.

devfile.yaml
```diff
@@ -11,30 +11,31 @@ schemaVersion: 2.1.0
...
       memoryLimit: 12Gi
+      memoryRequest: 512Mi
+      cpuRequest: 500m
       cpuLimit: 3500m
-      endpoints:
-        - exposure: public
-          name: dev
-          secure: true
-          protocol: http
-          targetPort: 8000
```
azatsarynnyy, line 24: With this endpoint, we get a handy `dev` item to quickly open a running dev instance of Che-Code:

image

Without this endpoint declared, one needs to use a notification to add a redirect manually:

image

as there's no endpoint, by default:

image

Why do we need to remove it?

devfile.yaml
```diff
@@ -11,30 +11,31 @@ schemaVersion: 2.1.0
...
+      mountSources: true
+
   - name: projects
     volume:
-      size: 3Gi
+      size: 7Gi
+
 commands:
-  - id: prepare
```
azatsarynnyy, line 29: If the purpose of adding `yarn-` prefix is for UI only, you can add [`exec.label` property](https://github.com/eclipse/che/issues/21625) with the descriptive command name.

devfile.yaml
vitaliy-guliy: This is it! 
I think why I am constantly asked to expose the port each time when I launch secondary instance of VS Code..

devfile.yaml
vitaliy-guliy: Done with that

devfile.yaml
```diff
@@ -11,44 +11,88 @@ schemaVersion: 2.1.0
...
+
   - name: dev
     container:
-      image: quay.io/che-incubator/che-code-dev:insiders
+      image: quay.io/che-incubator-pull-requests/che-code-dev:pr-253-dev-amd64
       memoryLimit: 12Gi
+      memoryRequest: 512Mi
+      cpuRequest: 500m
       cpuLimit: 3500m
+      mountSources: true
```
azatsarynnyy: Does it change the default behavior?
AFAIK, the sources should be mounted in the `dev` container without setting this explicitly.

build/dockerfiles/dev.Dockerfile
```diff
@@ -6,81 +6,49 @@
...
+USER 0
 
-FROM scratch
-COPY --from=ubi-micro-build /mnt/rootfs/ /
-ENV HOME=/home/che
-ENV NPM_CONFIG_PREFIX=/home/che/.npm-global
-ENV PATH /opt/nodejs/bin:/home/che/.npm-global/bin/:$PATH
-ENTRYPOINT /entrypoint.sh
-WORKDIR /projects
-RUN wget https://github.com/robbyrussell/oh-my-zsh/raw/master/tools/install.sh -O - | zsh && \
```
azatsarynnyy, line 70: Let's keep the "Oh My Zsh" framework installed. It makes the terminal in dev container much more dev-friendly )

BTW, when I open a terminal to the `dev` container, `bash` is running by default, but not `zsh` anymore, like we currently have in quay.io/che-incubator/che-code-dev:insiders. @vitaliy-guliy could please look at it as well?

build/dockerfiles/dev.Dockerfile
azatsarynnyy: It's called currently in quay.io/che-incubator/che-code-dev:insiders

azatsarynnyy: It's called currently in quay.io/che-incubator/che-code-dev:insiders

.github/workflows/pr-check-build-che-code-image.yaml
```diff
@@ -0,0 +1,88 @@
...
+      - '.che/*'
+      - '.rebase/*'
+
+jobs:
+  build-che-code:
+    runs-on: ubuntu-22.04
+    steps:
+
+      - name: Checkout che-code source code
+        uses: actions/checkout@v2
```
azatsarynnyy: Please, update it to `actions/checkout@v3`.
With `v2` we get a warning:

image

.github/workflows/pr-check-build-che-code-image.yaml
```diff
@@ -0,0 +1,88 @@
...
+
+      - name: Login to Quay.io
+        uses: docker/login-action@v2
+        with:
+          registry: quay.io
+          username: ${{ secrets.QUAY_PULL_REQUESTS_USERNAME }}
+          password: ${{ secrets.QUAY_PULL_REQUESTS_PASSWORD }}
+
+      - name: Compile linux-libc
```
azatsarynnyy: If we build libc and musl images [in parallel](https://github.com/che-incubator/che-code/blob/3aa3c9d2dc54ce332121fef82c72cf530fe8ca1e/.github/workflows/pr-check.yml#L22) the whole workflow will take about half as long to complete.

https://github.com/che-incubator/che-code/actions/runs/6111546375
This ^^ should take ~1 hour instead of ~2

azatsarynnyy (commented):
The commands for building the Che-Code image in the `dev` container with podman work as well 🎉   

image


Thanks @vitaliy-guliy!

devfile.yaml
```diff
@@ -11,44 +11,88 @@ schemaVersion: 2.1.0
...
+        REMOVE_TO="# Store the content of the result"
+        sed -i "/${REMOVE_FROM}/,/${REMOVE_TO}/d" build/dockerfiles/linux-libc.no-test.Dockerfile
+        podman build -f build/dockerfiles/linux-libc.no-test.Dockerfile -t linux-libc-amd64 .
+      group:
+        kind: run
+        isDefault: true
+
+  - id: podman-assemble-che-code
+    exec:
+      label: Build che-code docker image using Podman
```
azatsarynnyy: I suggest not to mention `docker` in the command label.

azatsarynnyy: @vitaliy-guliy do you see a use case when we would need to run just one of the commands `podman-compile-libc` or `podman-assemble-che-code`?

Can we introduce just a single command `Build Che-Code image` that would do everything that both these commands are doing? So, one command to get a final image with the assembly. WDYT?

build/dockerfiles/dev.Dockerfile
vitaliy-guliy: Do we consider https://github.com/robbyrussell/oh-my-zsh as trusted source?

If we compare popularity of `bash` and `zsh`, I think `bash` is more popular. Anyway, I didn't touch the configuration of the default shell. And it was even not configured for dev image.

devfile.yaml
vitaliy-guliy: Ok, then we need to review all the devfiles and cleanup them a bit.

.github/workflows/pr-check-build-che-code-image.yaml
vitaliy-guliy: I think here we need to have two separate actions.
It could happen I want to rebuild only `libc` or `musl`.
Let's leave them as it is, the user can always launch two tasks.

devfile.yaml
vitaliy-guliy: I think it's better to have them separately. If build of `che-code` image failed, you will have to compile `che-code` again.

build/dockerfiles/dev.Dockerfile
vitaliy-guliy: Btw, when I navigate to `https://github.com/robbyrussell/oh-my-zsh` I have been forwarded to `https://github.com/ohmyzsh/ohmyzsh`

.github/workflows/pr-check-build-che-code-image.yaml
vitaliy-guliy: Done

devfile.yaml
vitaliy-guliy: Label updated

.github/workflows/pr-check-build-che-code-image.yaml
azatsarynnyy: > It could happen I want to rebuild only libc or musl.

agree, and it's currently possible
because [the matrix strategy splits the `build` job](https://github.com/che-incubator/che-code/blob/dfeffd73c46760906aca16e25f306a33f4cdbb2e/.github/workflows/pr-check.yml#L17-L23) and makes it possible to have separate jobs for libc and musl variants:

image

But with the changes from this PR, we lose such a possibility because everything will run as a single job:

image

For the developer, it means that in case of any build failures whole workflow must be rerun from the beginning and the dev will need to wait for a couple of more hours.

build/dockerfiles/dev.Dockerfile
azatsarynnyy: > Do we consider https://github.com/robbyrussell/oh-my-zsh as trusted source?

Sure, it's a trusted source )
https://ohmyz.sh/
image

azatsarynnyy: > Btw, when I navigate to https://github.com/robbyrussell/oh-my-zsh I have been forwarded to https://github.com/ohmyzsh/ohmyzsh

Great, we can use the GitHub org link.

.github/workflows/pr-check-build-che-code-image.yaml
vitaliy-guliy: I was confused a bit and thought we were discussing devfile commands, will try to rework this action.

build/dockerfiles/dev.Dockerfile
vitaliy-guliy: I found it, `chsh -s $(which zsh) root` changes the default shell for the terminal

.github/workflows/pr-check-build-che-code-image.yaml
vitaliy-guliy: If we use matrix build, the intermediate result will be stored to artifacts like here
https://github.com/che-incubator/che-code/blob/main/.github/workflows/pr-check.yml#L38-L41

As I remember we discussed that and decided not to use artifacts to simplify the process.
Would you like to revert it back?

azatsarynnyy: Considering that storing intermediate containers as GH artifacts gives us such a nice dev UX, I'd keep using artifacts. By "nice dev UX" I mean:
- having parallel jobs with the possibility of rerunning just one of them in case of any build failure
- keeping the whole GH Workflow at least x2 faster

vitaliy-guliy: Done

build/dockerfiles/dev.Dockerfile
vitaliy-guliy: The command `chsh -s $(which zsh) root` changes the default shell for `root`. To change it for `user` we need to change it in the entrypoint [like this](https://github.com/che-incubator/che-code/blob/7.74.x/build/scripts/entrypoint-dev.sh#L16)

```
    echo "${USER_NAME:-che}:x:$(id -u):0:Eclipse Che:${HOME}:/bin/zsh" >> /etc/passwd
```

And I thought. What if we leave bash as default shell?
I've played a bit with `zsh` and did not find big benefits in comparing with `bash`, except it looks like new year tree.

azatsarynnyy: > And I thought. What if we leave bash as default shell?

I don't mind. We just need to make sure that ZSH still works as it's currently do in `che-incubator/che-code-dev:insiders`. The default shell can be easily configured through the `VS Code Default Profile`.


> except it looks like new year tree

Try to revert back the `.p10k.zsh` config
https://github.com/che-incubator/che-code/pull/253/files#diff-b3867b14785bf05eb0c09d7e9868fad2ceb8fcab8c3a12deefac0d5a485fc7b0L54
It customizes zsh, so it should look as minimalistic as it is currently in `che-incubator/che-code-dev:insiders`

.github/workflows/pr-check-build-che-code-image.yaml
```diff
@@ -0,0 +1,112 @@
...
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+#
+
+name: Build che-code images Pull Request check
```
azatsarynnyy: Can we make the workflow / job names shorter, so it takes less space and no information is hidden at the end of each line?

image

azatsarynnyy: We can reuse the same names as we currently have:

image

We're in che-code repository, so adding che-code for each name doesn't bring much value. WDYT?

.github/workflows/pr-check-build-che-code-image.yaml
azatsarynnyy: Cool! Thanks @vitaliy-guliy 👍

build/dockerfiles/dev.Dockerfile
vitaliy-guliy: Done

vitaliy-guliy:
@azatsarynnyy I cannot say for sure, but seems the terminal looks like as it was before

Screenshot from 2023-10-04 16-27-32

build/dockerfiles/dev.Dockerfile
l0rd: The benefit of using UDI as the base image is that all UDI layers are already in the nodes of the cluster (especially if kube-image-puller is configured, so it should be really fast to pull images based on UDI on a kube cluster.

build/dockerfiles/dev.Dockerfile
```diff
@@ -6,68 +6,34 @@
...
+
+RUN dnf -y install libsecret libX11-devel libxkbcommon \
+    "https://rpmfind.net/linux/centos/8-stream/BaseOS/x86_64/os/Packages/libsecret-devel-0.18.6-1.el8.x86_64.rpm" \
+    "https://rpmfind.net/linux/centos/8-stream/AppStream/x86_64/os/Packages/libxkbfile-1.1.0-1.el8.x86_64.rpm" \
+    "https://rpmfind.net/linux/centos/8-stream/PowerTools/x86_64/os/Packages/libxkbfile-devel-1.1.0-1.el8.x86_64.rpm" \
+    "https://rpmfind.net/linux/centos/8-stream/BaseOS/x86_64/os/Packages/zsh-5.5.1-6.el8_1.2.x86_64.rpm" \
+    util-linux-user
+
+# cleanup dnf cache
+RUN dnf -y clean all --enablerepo='*'
```
l0rd: To reduce the image size it's better to install and clean in the same command otherwise we are going to commit a layer with all the temporary files and then a second layer with the clean up.

```suggestion
RUN dnf -y install libsecret libX11-devel libxkbcommon \
    "https://rpmfind.net/linux/centos/8-stream/BaseOS/x86_64/os/Packages/libsecret-devel-0.18.6-1.el8.x86_64.rpm" \
    "https://rpmfind.net/linux/centos/8-stream/AppStream/x86_64/os/Packages/libxkbfile-1.1.0-1.el8.x86_64.rpm" \
    "https://rpmfind.net/linux/centos/8-stream/PowerTools/x86_64/os/Packages/libxkbfile-devel-1.1.0-1.el8.x86_64.rpm" \
    "https://rpmfind.net/linux/centos/8-stream/BaseOS/x86_64/os/Packages/zsh-5.5.1-6.el8_1.2.x86_64.rpm" \
    util-linux-user && \
    dnf -y clean all --enablerepo='*'
```

build/dockerfiles/dev.Dockerfile
```diff
@@ -6,68 +6,34 @@
...
+
+# Create `/etc/containers/containers.conf` configuration file
+# to increase amount of opened files
+#
+# [containers]
+# default_ulimits = [
+#  "nofile=65535:65535",
+# ]
+
+RUN (echo '[containers]'; echo 'default_ulimits = ['; echo ' "nofile=65535:65535",'; echo ']') >> /etc/containers/containers.conf
```
l0rd: This may be something worth porting to the UDI itself

devfile.yaml
```diff
@@ -11,44 +11,87 @@ schemaVersion: 2.1.0
...
+      commandLine: |
+        echo "Compile libc mod"
+        cp -f build/dockerfiles/linux-libc.Dockerfile build/dockerfiles/linux-libc.no-test.Dockerfile
+        REMOVE_FROM="### Testing"
+        REMOVE_TO="# Store the content of the result"
+        sed -i "/${REMOVE_FROM}/,/${REMOVE_TO}/d" build/dockerfiles/linux-libc.no-test.Dockerfile
+        podman build -f build/dockerfiles/linux-libc.no-test.Dockerfile -t linux-libc-amd64 .
+      group:
+        kind: run
+        isDefault: true
```
l0rd: There are 3 commands of kind `run` with `isDefault` set to `true`. But there should be only one default command per kind.

build/dockerfiles/dev.Dockerfile
l0rd: But this should not block this PR though.

azatsarynnyy (approved):
LGTM 👍 
Thanks Vitaliy for your cooperation in addressing all my comments 🙂

.github/workflows/pr-check-build-che-code-image.yaml
vitaliy-guliy: Done

build/dockerfiles/dev.Dockerfile
vitaliy-guliy: Solved

devfile.yaml
vitaliy-guliy: Done

build/dockerfiles/dev.Dockerfile
vitaliy-guliy: Done, the file has been added to the UDI

devfile.yaml
vitaliy-guliy: Done

PR #256, opened 2023-07-07
Move part of logic configuring che-code to javascript instead of keeping it in bash

vitaliy-guliy:
### What does this PR do?

Adds a simple Typescript project that is compiled to javascript and injected into the editor container.

What does it do? Exactly as the corresponding blocks in the entrypoint:

- configures OpenVSIX registry. If OPENVSX_REGISTRY_URL environment variable is set and not empty, applies it as a registry URL. If OPENVSX_REGISTRY_URL is empty, uses CHE_PLUGIN_REGISTRY_URL as a registry URL.

- if WEBVIEW_LOCAL_RESOURCES environment variable is set to `true`, configures webview static resources to be loaded from the local host (instead of getting them from CDN). It makes webview working when the browser is behind a proxy.

- Creates a bundle certificate and initializes NODE_EXTRA_CERTIFICATE environment variable. The bundle certificate contains content of `/tmp/che/secret/ca.crt` and all found certificates in `/public-certs`.

- Generates `/projects/.code-workspace` file containing all the projects taken from devworkspace flattened devfile.

- Prepares all necessary parameters and launches VS Code.

https://github.com/eclipse/che/issues/22280

### How to test this PR?

**Important:** it is not possible to build PR-based images due to https://github.com/eclipse/che/issues/22390

On the dogfooding instance (due to 12Gb for dev container), create a workspace with https://github.com/che-incubator/che-code/tree/che-code-launcher-test

You should have the workspace with three projects opened.

Screenshot from 2023-07-07 13-29-19


nallikaea:
Hi, @vitaliy-guliy 
It is possible to use replace() method instead of "sed"  using something like this:
```
fs.readFileSync("out/vs/workbench/workbench.web.main.js", 'utf8', function (err, data) {
  if (err) {
    return console.log(err);
  }
  const result = data.replace(/substringToReplaceAsRegEx/g, 'replacement');
  fs.writeFileSync("out/vs/workbench/workbench.web.main.js", result, 'utf8', function (err) {
     if (err) return console.log(err);
  });
});
```
For JSON files it can be another solution:
```
    const productJSON = fs.readFileSync("product.json", "utf-8");
    const data = JSON.parse(productJSON);
    data.itemUrl = ""${openvsxURL}/item";
    fs.writeFileSync("product.json", JSON.stringify(data), "utf-8");
```
[1] https://stackoverflow.com/questions/14177087/replace-a-string-in-a-file-with-nodejs
[2] https://stackoverflow.com/questions/52323813/replacing-a-string-in-javascript-while-keeping-json-file-intact

vitaliy-guliy:
> Hi, @vitaliy-guliy It is possible to use replace() method instead of "sed" using something like this:
> 

Yes, I'm thinking about this as well. Using `sed` was just a copypasting the pieces of code and it was a simpler way to transfer the functionality without worrying that I broke something.
Thanks!

vitaliy-guliy:
@nallikaea @musienko-maxim could you take a look as well?

build/dockerfiles/linux-libc.Dockerfile
```diff
@@ -120,6 +126,18 @@ RUN [[ $(uname -m) == "x86_64" ]] && NODE_ARCH=$(echo "console.log(process.arch)
...
+#
+# Copy VS Code launcher to the container
+#
+#########################################################
+COPY launcher /checode-launcher
+WORKDIR /checode-launcher
+RUN yarn \
+    && mkdir /checode/launcher \
+    && cp -r out/src/*.js /checode/launcher \
+    && chgrp -R 0 /checode && chmod -R g+rwX /checode
```
RomanNikitenko, line 139: I guess the same changes are required for the downstream

launcher/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "clean": "rimraf lib",
+    "build": "yarn run format && yarn run compile && yarn run lint && yarn run test",
+    "compile": "tsc --project .",
+    "format": "prettier --check '{src,tests}/**/*.ts' package.json",
+    "format:fix": "prettier --write '{src,tests}/**/*.ts' package.json",
+    "lint": "eslint --cache=true --no-error-on-unmatched-pattern=true '{src,tests}/(!model|**)/*.ts'",
+    "lint:fix": "eslint --fix --cache=true --no-error-on-unmatched-pattern=true \"{src,tests}/(!model|**)/*.{ts,tsx}\"",
+    "test": "jest --forceExit",
+    "watch": "tsc -w",
+    "publish:next": "yarn publish  --registry=https://registry.npmjs.org/ --no-git-tag-version --new-version 0.0.1-\"$(date +%s)\""
```
RomanNikitenko: Are we really going to publish the launcher?

launcher/src/flattened-devfile.ts
```diff
@@ -0,0 +1,96 @@
...
+  attributes: KeyValue;
+}
+
+export class FlattenedDevfile {
+  private devfile: Devfile;
+
+  async getDevfile(): Promise<Devfile> {
+    if (!env.DEVWORKSPACE_FLATTENED_DEVFILE) {
+      throw new Error(
+        "  > Unable to find flattened dewworkspace file, env.DEVWORKSPACE_FLATTENED_DEVFILE is not set"
```
RomanNikitenko: dewworkspace => de**v**workspace

launcher/src/fs-extra.ts
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ ***********************************************************************/
+
+/**
+ * This is a wrapper to comfortable using `fs` package like `fs-extra`.
+ * Why do we need to add the wrapper:
+ *   - it allows to get rid of using of `fs-extra` package
+ *   - it allows to use common `node_modules` directory for launcher and VS Code
+ *   - it simplifies writing tests and alows to easily mock of this module
```
RomanNikitenko: alows => allows

RomanNikitenko: also, maybe `of` is not needed here, but I'm not English expert :-)

launcher/src/openvsix-registry.ts
```diff
@@ -0,0 +1,98 @@
...
+      return;
+    }
+
+    try {
+      let openvsxURL;
+      if (env.OPENVSX_REGISTRY_URL) {
+        console.log(
+          `  > env.OPENVSX_REGISTRY_URL set to ${env.OPENVSX_REGISTRY_URL}`
+        );
+        openvsxURL = `${env.OPENVSX_REGISTRY_URL}/vscode`;
```
RomanNikitenko: I think it makes sense to check what value we get from the env variable
let's say `https://open-vsx.org/` comes from the env variable
then there is a problem: URL contains double `/`:  `https://open-vsx.org//vscode`

I remember someone reported such problem...

launcher/package-lock.json
```diff
@@ -0,0 +1,9963 @@
+{
```
RomanNikitenko: is `package-lock.json` file required?
I mean - VS Code uses `yarn.lock` files, Che-Code uses `yarn.lock` files for own extensions as well...

launcher/package.json
vitaliy-guliy: It was copypasted from another project, thanks!

launcher/src/openvsix-registry.ts
vitaliy-guliy: The same for CHE_PLUGIN_REGISTRY_URL.
I've added the verification for that, thanks!

launcher/package-lock.json
vitaliy-guliy: I'm thinking how was this file appear in the project...
Thanks.

musienko-maxim:
1. The functionality was checked on Devspace3.7.1+Openshift 4.12. So all worked properly. The errors with the self-signed certificate were expected.

RomanNikitenko (approved):

RomanNikitenko:
I think https://github.com/eclipse/che/issues/22345 can be closed as well after merging current PR.
thanks Vitaliy!

azatsarynnyy (approved):
I've tested these changes and confirm that it fixes the original problem described in https://github.com/eclipse/che/issues/22280.
Thanks @vitaliy-guliy for simplifying the entrypoint script by moving the related logic to TS code :+1:

build/dockerfiles/linux-libc.Dockerfile
RomanNikitenko: @vitaliy-guliy 
the current PR was merged
are you going to provide similar changes to the downstream?

RomanNikitenko: I think downstream already broken: 
- entrypoint in the downstream already contains your changes: https://github.com/redhat-developer/devspaces-images/blob/devspaces-3-rhel-8/devspaces-code/build/scripts/entrypoint-volume.sh#L50
- but the dockerfile is not adapted...

vitaliy-guliy: I suppose the downstream was updated automatically
https://github.com/redhat-developer/devspaces-images/tree/devspaces-3-rhel-8/devspaces-code

RomanNikitenko: Did you check the corresponding dockerfile?
I believe the similar changes are required for the dockerfile...

vitaliy-guliy: Done https://github.com/redhat-developer/devspaces-images/pull/491

PR #258, opened 2023-07-11
Nodejs from sources

RomanNikitenko:


RomanNikitenko:
Another approach was applied within PR: https://github.com/che-incubator/che-code/pull/324

PR #261, opened 2023-07-26
fix: install missing modules for kerberos authentication

vitaliy-guliy:
### What does this PR do?
Fixes build of `linux-musl-amd64` image by adding missing `kerberos` library. 

The dependency was added by this commit https://github.com/microsoft/vscode/commit/6ae441b56fc9ca54b53a4d38810619f326b56bec

I did nothing for `linux-libc-amd64` as we do not have issues with it right now, but remember for future that it may require the dependency as well.

https://github.com/eclipse/che/issues/22390

### How to test this PR?
...


RomanNikitenko (approved):

RomanNikitenko:
@vitaliy-guliy 
thank you very much for the quick fix!

PR #266, opened 2023-08-09
Name the rebase rules correctly

azatsarynnyy:
### What does this PR do?
Fixes naming of the rebase rules.

addresses https://github.com/che-incubator/che-code/pull/252#discussion_r1287962785
https://github.com/eclipse/che/issues/22428


RomanNikitenko (commented):
@azatsarynnyy 
Did you test the changes?

it fails in another place for me...
the error is still related to your git-rebase logic for mangler

azatsarynnyy:
@RomanNikitenko I added one more patch (in this PR) to the rebase settings. It should work well now.

RomanNikitenko:
> @RomanNikitenko I added one more patch (in this PR) to the rebase settings. It should work well now.

Actually from my point of view automatic git-rebase for the mangler is not complete.
Every change in the `che-code/code/build/lib/mangle/index.js` should lead to changes for encoded data at the end of the file, see in the upstream for example: https://github.com/microsoft/vscode/commit/8f4385d10c0a5fc5161c71ae1572b238dc3a8e5a#diff-d2f03589ad1b879f58cffaf7b20bc9906b73f2c901010287ea258c61b76d2157L665

Instead of that I have uncommited changes for the `che-code/code/build/lib/mangle/index.js` file after rebuild the corresponding module - it means that encoded data doesn't reflect actual data in the file.

image

So, maybe it makes sense to add the corresponding logic to the `apply_mangle_index_js_changes() ` method.
WDYT?

azatsarynnyy:
Thank you @RomanNikitenko for testing this.
I'm not sure I understand how to get the state from the screenshot above. We do [`git checkout --theirs ...`](https://github.com/che-incubator/che-code/blob/5ab3b5ac27288b9191ae859061c546a514256d4b/rebase.sh#L176), so when I perform:
```
./rebase.sh
yarn
```
I get:
```
On branch reb-fix
Your branch is ahead of 'origin/reb-fix' by 158 commits.
  (use "git push" to publish your local commits)

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   code/extensions/che-github-authentication/yarn.lock
	modified:   code/test/smoke/yarn.lock
```

Anyway, the answer to your question, I'd say, depends on the reason why you need those encoded data and how you use them.

RomanNikitenko:
>  I'd say, depends on the reason why you need those encoded data and how you use them.

Sorry, I'm afraid I don't understand your question - why do you ask how **I** am going to use that data. 
These changes are going be merged **not to my fork**, but to the **community** version :-)

Could you clarify:
- Do you mean that VS Code doesn't need those data at all? 
or 
- VS Code do need but Che-Code doesn't?

RomanNikitenko:
@azatsarynnyy 
I've pushed a fix for the https://github.com/che-incubator/che-code/pull/266#issuecomment-1672810036

How to test my changes:
- `./rebase.sh` **without** my changes => `cd code/build && yarn compile` => as result **you have changes** in the js file
- `./rebase.sh` **with** my changes => `cd code/build && yarn compile` => **there are no changes** in the js file

RomanNikitenko (approved):

PR #273, opened 2023-09-08
chore: browser database name should contain a valid DevWorkspace identifier

vitaliy-guliy:
### What does this PR do?
Uses a real workspace identifier taken from env.DEVWORKSPACE_ID to form the name of browser-based database.

https://github.com/eclipse/che/issues/22338

### How to test this PR?

- When you are on the dashboard, open browser development console, go to `Application`. In the `Storage` section below expand `IndexedDB` and remove all the databases
- Create a workspace with https://github.com/che-incubator/che-code/tree/vscode-web-state-db-identifier-test
- open a terminal, show a read `DEVWORKSPACE_ID`

```
$ export | grep DEVWORKSPACE_ID
```
- [1] ensure there is a database containing the ID at the end

- go to the dasboard, delete workspace, create a workspace using the same URI
- [2] check the workspace ID and check, that there is another database with this ID has been created

Browse URL must be the same for both workspaces.

Screenshot from 2023-09-08 11-06-41


code/src/vs/workbench/services/workspaces/browser/che/devWorkspaceId.ts
```diff
@@ -0,0 +1,24 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ ***********************************************************************/
+/* eslint-disable header/header */
+
+// !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+// NOTE: DO NOT CHANGE. Launcher updates this constant on workspace startup
+// !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+const DEVWORKSPACE = "https://{{che-cluster}}.{{host}}/{{namespace}}/{{workspace-name}}/{{port}}/";
+
+export function getDevWorkspceId(): string | undefined {
```
azatsarynnyy: `getDevWorkspceId` > `getDevWorkspaceId`

code/src/vs/workbench/services/workspaces/browser/workspaces.ts
```diff
@@ -30,5 +31,5 @@ export function getSingleFolderWorkspaceIdentifier(folderUri: URI): ISingleFolde
 }
 
 function getWorkspaceId(uri: URI): string {
-	return hash(window.location.href + uri.toString()).toString(16);
+	return getDevWorkspceId() || hash(uri.toString()).toString(16);
```
azatsarynnyy: I noticed that using only DevWorkspaceID as a VS Code's Workspace ID will bring another problem.

Let's consider the situation when the user has a DevWorkspace with the Che-Code sources cloned, but she wants to open just a subfolder, e.g.: `launcher`. In such a case, these two **different** VS Code Workspaces (`che-code` and `che-code/launcher`) will share the same IndexedDB. This means the user will get the same stored state of VS Code (panels layout, opened files, etc.) in both workspaces which is not expected, probably.
WDYT @vitaliy-guliy ?

If we decide we should avoid such a situation I suggest generating a hash value from the DevWorkspace ID as well: `hash(getDevWorkspceId() + uri.toString()).toString(16)`

vitaliy-guliy: Ok, in this case we will definitely have the unique ID :grinning:

azatsarynnyy (approved):

PR #260, opened 2023-07-24
feat: Add Device Authentication flow support

RomanNikitenko:
### What does this PR do?
- Adds `Github:  Device Authentication` command
- Adds `Github: Remove Device Authentication Token` command
- Stores `Device Authentication` token to a separate secret
- Resolves a problem when there are few tokens in the `.git-credential/credentials` file, see https://github.com/eclipse/che/issues/22433.
 

https://github.com/che-incubator/che-code/assets/5676062/2fc017cd-bc56-466a-93e3-fecc43c679c8

- Adds a logger to the `Eclipse Che API` and `Che Github Authentication` extensions, so logs are available in the corresponding channels of the `Output` view
image

### What issues does this PR fix?

https://github.com/eclipse/che/issues/22139

#### Use case 1 - an extension authorization
- Start a workspace 
- Install an extension for testing that requires `Device Authentication` flow for authorization
- `F1 => Device Authentication` => follow the flow
- Reload the window to clean up cached sessions
- Sign In
- Try to use the extension

#### Use case 2 - Git related operations should work correctly when a Device Authentication token is used 
- Steps from the use case 1
- `F1` => `Git Clone` => follow the flow
- check that a project is cloned successfully

#### Use case 3 - Git related operations should work correctly after removing a Device Authentication token 
- Steps from the use case 1
- `F1` => `Github: Remove Device Authentication Token`
- Click the button `Accounts` on the left panel => `Github Sign Out`
- Try to test use case 2 - the `Git Clone` command should work correctly, using a token from the `.git-credential/credentials` file (if it's present)

#### Use case 4 - the changes should fix `Error: Invalid character in header content` problem when there are few tokens
- Create a workspace from the `gitlab` repo - the `gitlab` token should be added to the `.git-credential/credentials` file
- Stop the workspace ^.
- Create a workspace from the `github` repo - the `github` token should be added to the `.git-credential/credentials` file
- Check that `.git-credential/credentials` file contains both tokens
- `F1` => `Git Clone` => follow the flow
- check that a project is cloned successfully
- so, the following issue should be fixed https://github.com/eclipse/che/issues/22433

#### Use case 5 - Device Authentication flow can be used for authentication when there is no git-credential secrets/tokens
- Delete all `git-credential` secrets
- Start a workspace without authentication on the Github at the starting workspace step
- Check that the `.git-credential/credentials` file is empty
- `F1` => `Git Clone` => follow the flow => you should get an error - there is no token 
[...]

PR #277, opened 2023-09-28
fix: Allow GitHub clone when credentials file contains several tokens

vitaliy-guliy:
### What does this PR do?
Allows to clone from GitHub when `/.git-credentials/credentials` contains several different tokens.

Screenshot from 2023-09-28 20-55-31

https://github.com/eclipse/che/issues/22433

### How to test this PR?
- Create a workspace from this branch ( due to resource limitations works for dogfooding instance only )
- Ensure you have several tokens in `/.git-credentials/credentials`.
  If no, delete this workspace, create any workspace using GitLab repository, then stop created workspace and create a workspace from this branch again
- Run `devfile: prepare` task to install node dependencies
- Run `devfile: build` task and then run `devfile: run` tack
- A new browser tab should be opened, try GitHub clone

---
The code change were made using Che only :sunglasses:

### DIFF
--- a/code/extensions/che-api/src/impl/github-service-impl.ts
+++ b/code/extensions/che-api/src/impl/github-service-impl.ts
@@ -16,6 +16,8 @@ import { AxiosInstance } from 'axios';
 import * as fs from 'fs-extra';
 import * as path from 'path';
 
+const GITHUB_GET_AUTHENTICATED_USER = 'https://api.github.com/user';
+
 @injectable()
 export class GithubServiceImpl implements GithubService {
     private readonly token: string | undefined;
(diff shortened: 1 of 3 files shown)


code/extensions/che-api/tests/github-service-impl.spec.ts
```diff
@@ -12,59 +12,83 @@
 import 'reflect-metadata';
```
RomanNikitenko, line 12: unfortunately tests are not integrated in the build process
so, I would say they are run never: https://github.com/eclipse/che/issues/22213
some of them already broken: https://github.com/che-incubator/che-code/pull/227#issuecomment-1548387927
the only way to run them at the moment - manually...

vitaliy-guliy:
I wanted to check the case when `/.git-credentials/credentials` contains several keys for github and only one of them is actual. But when I played with secrets, the valid key was always first in the file.
From the other side, in the service I used dummy value as a token and noticed it never used. So need to review the idea of reading the credentials file at all.

```
    constructor(@inject(Symbol.for('AxiosInstance')) private readonly axiosInstance: AxiosInstance) {
        this.token = 'anykey';
    }
```

I leave changes as it is till future improvements.

code/extensions/che-api/src/impl/github-service-impl.ts
```diff
@@ -16,40 +16,50 @@ import { AxiosInstance } from 'axios';
...
 
     constructor(@inject(Symbol.for('AxiosInstance')) private readonly axiosInstance: AxiosInstance) {
         const credentialsPath = path.resolve('/.git-credentials', 'credentials');
         if (fs.existsSync(credentialsPath)) {
-            const token = fs.readFileSync(credentialsPath).toString();
-            this.token = token.substring(token.lastIndexOf(':') + 1, token.indexOf('@'));
+            const content = fs.readFileSync(credentialsPath).toString();
+
+            // Since this service is specific for github.com, we need to find only one corresponding token
+            const regex = /(http|https):\/\/oauth2:(\d|\S+)@github.com$/mg;
```
RomanNikitenko, line 31: we should test it for the case when custom git server is used
I mean - before your changes a token was read this way:
```
this.token = token.substring(token.lastIndexOf(':') + 1, token.indexOf('@'));
```
As far as I understand, now it depends on the `github.com` pattern, so it can bring a regression for the custom git server use case

vitaliy-guliy: I suppose we are talking about GitHub, which is `github.com`

vitaliy-guliy: Anyway, I was surprised it works with token 'anykey'.

vitaliy-guliy:
This PR is outdated,
the original issue solved by https://github.com/che-incubator/che-code/pull/260

PR #301, opened 2023-11-14
[ports-plugin] don't prompt user to redirect endpoints with `exposure: none`

### LINKED ISSUE #22655: Don't ask to add a redirect for endpoints with non `public` exposure

Containers endpoints can have an exposure `public`, `internal` or `none` (see the [documentation](https://devfile.io/docs/2.2.1/devfile-schema#components-container-endpoints-exposure)).  

If a developer specifies that an endpoint in a devfile has exposure `internal` or `none`

```yaml
components:
  - name: tools
    container:
      image: quay.io/devfile/universal-developer-image:ubi8-latest
      endpoints:
        - exposure: none
          name: debug
          protocol: tcp
          targetPort: 5005
``` 

we are still proposing to add a redirect:

image

We should not show a prompt for endpoints that have exposure other than `public`.

amisevsk:
### What does this PR do?
Updates how endpoints are handled in the ports plugin:

* If a process starts listening internally (i.e. on localhost) to a port that matches an endpoint with `exposure: none`, do **not** prompt the user to redirect this port
* If a process starts listening internally (i.e. on localhost) to a port that matches an endpoint with `exposure: public`, change the redirect prompt to mention this specifically, as it may indicate unexpected behavior in the editor (redirecting a process rather than serving it over the defined endpoint in the devfile)

This PR also removes some old JWT handling code, since we don't use JWT anymore.

Closes https://github.com/eclipse/che/issues/22655

### How to test this PR?
The easiest way to test this PR is via this devfile https://gist.github.com/amisevsk/2f11f0e282b93dc6413b0dd360437a7b and the following simple script (`test.js`):
```javascript
var http = require('http');
http.createServer(function (req, res) {}).listen(1110); // public endpoint, listening on all addresses
http.createServer(function (req, res) {}).listen(1111); // internal endpoint, listening on all addresses
http.createServer(function (req, res) {}).listen(1112); // none endpoint, listening on all addresses

http.createServer(function (req, res) {}).listen(2220, "localhost"); // public endpoint, listening on localhost
http.createServer(function (req, res) {}).listen(2221, "localhost"); // internal endpoint, listening on localhost
http.createServer(function (req, res) {}).listen(2222, "localhost"); // none endpoint, listening on localhost
```

The devfile has six endpoints:
* `1110`, `1111`, and `1112` are `public`, `internal`, and `none` endpoints (in that order) for testing "listen on 0.0.0.0"
* `2220`, `2221`, and `2222` are `public`, `internal`, and `none` endpoints (in that order) for testing "listen on localhost"

If you run the `test.js` script, you should see:
* Two prompts -- one for the public endpoint and one warning/prompt for the public-internal endpoint
    

* Four logs in the ports plugins output pane:
    ```
    Endpoint test-none on port 1112 is defined as exposure: none. Do not prompt to open it.
    Endpoint test-internal on port 1111 is defined as exposure: internal. Do not prompt to open it.
    Endpoint test-local-internal on port 2221 is defined as exposure: internal. Do not prompt to open it.
    Endpoint test-local-none on port 2222 is defined as exposure: none. Do not prompt to open it.
    ```

### DIFF
--- a/code/extensions/che-port/src/devfile-handler-devworkspace-impl.ts
+++ b/code/extensions/che-port/src/devfile-handler-devworkspace-impl.ts
@@ -84,27 +84,6 @@ export class DevWorkspaceDevfileHandlerImpl implements DevfileHandler {
       } as Endpoint;
     });
 
-    // Add private JWT proxy ports
-    const jwtProxyEnv: string[] = Object.keys(process.env).filter(key =>
-      key.includes('_JWTPROXY_SERVICE_PORT_SERVER_')
-    );
-    jwtProxyEnv.forEach((key, index) => {
-      const value = process.env[key]!.toLocaleLowerCase() || '';
-      const port = parseInt(value);
-      if (!isNaN(port)) {
-        const endpoint: Endpoint = {
-          name: `jwt-proxy-${index + 1}`,
-          exposure: EndpointExposure.FROM_DEVFILE_PRIVATE,
-          url: '',
-          targetPort: port,
-          protocol: 'tcp',
-          type: 'jwt-proxy',
-          category: EndpointCategory.PLUGINS,
-        };
-        endpoints.push(endpoint);
-      }
-    });
-
     return endpoints;
   }
 }
--- a/code/extensions/che-port/src/ports-plugin.ts
+++ b/code/extensions/che-port/src/ports-plugin.ts
@@ -172,11 +172,19 @@ export class PortsPlugin {
     // check now if the port is in workspace definition ?
     const matchingEndpoint = this.devfileEndpoints.find(endpoint => endpoint.targetPort === port.portNumber);
 
-    if (matchingEndpoint && matchingEndpoint.exposure === EndpointExposure.FROM_DEVFILE_PRIVATE) {
-      this.outputChannel.appendLine(
-        `Endpoint ${matchingEndpoint.name} on port ${matchingEndpoint.targetPort} is defined as Private. Do not prompt to open it.`
-      );
-      return;
+    if (matchingEndpoint) {
+      if (matchingEndpoint.exposure === EndpointExposure.FROM_DEVFILE_PRIVATE) {
+        this.outputChannel.appendLine(
+          `Endpoint ${matchingEndpoint.name} on port ${matchingEndpoint.targetPort} is defined as exposure: internal. Do not prompt to open it.`
+        );
+        return;
+      }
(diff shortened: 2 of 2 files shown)


amisevsk:
The smoke test failure is due to a current quay.io outage

azatsarynnyy (approved):
I tested the patch with my [test project](https://github.com/azatsarynnyy/java-spring-petclinic/tree/ports) and I can confirm that it works as described above.
I just had to modify the last line in the test.js script to listen on 2223 instead of 2222 according to the edpoints declared in the provided test Devfile.
Thanks @amisevsk for your PR and for removing the obsolete code.

azatsarynnyy:
I rerun the smoke tests.

amisevsk:
> test.js script to listen on 2223 instead of 2222 according to the edpoints declared in the provided test Devfile.

Ah yeah, I fixed that in my tests but I fixed it in the DevWorkspace and forgot to update the devfile. I will update the sample devfile :+1:

vitaliy-guliy:
I confused a bit. Why do we have the first prompt if we have the same definition for 2220 and 1110?
```
        - exposure: public
          name: test-public
          secure: true
          protocol: http
          targetPort: 1110
          
        - exposure: public
          name: test-local-public
          secure: true
          protocol: http
          targetPort: 2220
```

Anyway, I haven't played yet with http server bind to a specific ip/host.
But at the first glance it's strange.

vitaliy-guliy (approved):
Tested using @azatsarynnyy sample, works as described.

amisevsk:
> Why do we have the first prompt if we have the same definition for 2220 and 1110?

The first prompt is because the server that's listening on port 2220 is only listening on localhost (instead of 0.0.0.0), and so would be inaccessible from outside the container. To connect to this server, we would need an in-container proxy (which is what Code will do if you accept the prompt) -- the route/ingress created for port 2220 will not be able to connect.

Previously, the user would see the same prompt as for other non-exposed endpoints (e.g. if you listen on port 9999), where Code is asking if you want it to proxy connections to 9999 within the container. I changed the prompt to point out the potential issue, since the public endpoint suggests the user wants to use that endpoint's URL to connect instead of a redirect route/ingress.

code-port-plugin-flow

amisevsk:
For reference, this is what the prompts look like in the current main branch:



The first one in this screenshot is due to https://github.com/eclipse/che/issues/22655. The second one is the generic "your process is listening on localhost instead of 0.0.0.0", which is changed in this PR to reflect that an endpoint exists and isn't being used.

PR #307, opened 2023-11-20
ripgrep 1.15.6 -> 1.15.7

Sachin-Itagi:
### What does this PR do?
This PR bumps the version of ripgrep from 1.15.6 to the latest version 1.15.7

Fixes the issue related to search function for Power.
https://issues.redhat.com/browse/CRW-4938

### How to test this PR?
In the container for Power, `/checode-linux-libc/node_modules/@vscode/ripgrep/bin/rg --version` shouldn't give an error.

### DIFF
--- a/code/package.json
+++ b/code/package.json
@@ -72,7 +72,7 @@
     "@vscode/iconv-lite-umd": "0.7.0",
     "@vscode/policy-watcher": "^1.1.4",
     "@vscode/proxy-agent": "^0.18.1",
-    "@vscode/ripgrep": "^1.15.6",
+    "@vscode/ripgrep": "^1.15.7",
     "@vscode/spdlog": "^0.13.12",
     "@vscode/sqlite3": "5.1.6-vscode",
     "@vscode/sudo-prompt": "9.3.1",
--- a/code/remote/package.json
+++ b/code/remote/package.json
@@ -8,7 +8,7 @@
     "@parcel/watcher": "2.1.0",
     "@vscode/iconv-lite-umd": "0.7.0",
     "@vscode/proxy-agent": "^0.18.1",
-    "@vscode/ripgrep": "^1.15.6",
+    "@vscode/ripgrep": "^1.15.7",
     "@vscode/spdlog": "^0.13.12",
     "@vscode/vscode-languagedetection": "1.0.21",
     "@vscode/windows-process-tree": "^0.5.0",
--- a/code/remote/yarn.lock
+++ b/code/remote/yarn.lock
@@ -72,10 +72,10 @@
   optionalDependencies:
     "@vscode/windows-ca-certs" "^0.3.1"
 
-"@vscode/ripgrep@^1.15.6":
-  version "1.15.6"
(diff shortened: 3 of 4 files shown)


ibuziuk (approved):

code/package.json
```diff
@@ -72,7 +72,7 @@
     "@vscode/iconv-lite-umd": "0.7.0",
     "@vscode/policy-watcher": "^1.1.4",
     "@vscode/proxy-agent": "^0.18.1",
-    "@vscode/ripgrep": "^1.15.6",
+    "@vscode/ripgrep": "^1.15.7",
```
RomanNikitenko, line 75: as far as I understand the upstream still uses `1.15.6`version: https://github.com/microsoft/vscode/blob/047cf79f5d10df18fcf03dbdffe84c107d5c8cbd/package.json#L75

I'm afraid the current change will be overwritten by the upstream changes at the next synchronization Che-Code with the VS Code - so in few hours...

RomanNikitenko: to avoid it the corresponding part should be provided to the `.rebase` folder

nickboldt: the last 2 merges / rebases from upsteam didn't touch the code/yarn.lock or code/remote/yarn.lock:

https://github.com/che-incubator/che-code/commit/23ace8433f67a10f754c3cf5242668e02689cf99

https://github.com/che-incubator/che-code/commit/61ab6eeeb4bdfdf037ab63afe17087935694b7d5#diff-097ceb3f628ecb473c665dbf54d6571f0b282e6a6f91486bc0c06b74960c243d

But this should be rebased against 7.77.x and merged there, where rebasing from upstream doesn't happen. Correct @RomanNikitenko  ?

nickboldt, line 75: Also we should get https://github.com/microsoft/vscode-ripgrep/pull/54 merged and THEN bump this to 1.15.8

nickboldt (changes requested):

code/package.json
RomanNikitenko: > But this should be rebased against 7.77.x and merged there, where rebasing from upstream doesn't happen.

correct!

I think we don't need this change for the main branch as VS Code is going to update the version this week.

levivic: ping @Sachin-Itagi , https://github.com/microsoft/vscode-ripgrep/pull/54 is merged now.

RomanNikitenko: the version is going to be updated in the upstream soon https://github.com/microsoft/vscode/pull/198880/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R75

RomanNikitenko:
@Sachin-Itagi 
I fetched the latest changes from the upstream, so now Che-Code has `@vscode/ripgrep": "^1.15.9`.

Can we close the PR?

nickboldt:
Did you regen the yarn.lock? May need that for downstream cachito stuff that @mkuznyetsov is working on in [CRW-3160](https://issues.redhat.com//browse/CRW-3160)

If so then yes, this PR is obsolete and should be closed.

PR #310, opened 2023-11-21
Add remote settings.json to injector and copy it on start.

### LINKED ISSUE #22653: Default path when cloning a git repository using VS Code command palette should be $PROJECTS_ROOT

When cloning a Git repository from VS Code command palette, the default folder is $HOME and not $PROJECTS_ROOT. If the user accept the default path, the changes to the source code won't be persisted.

image

The default path should be $PROJECTS_ROOT.

_No response_

### Additional context

_No response_

amisevsk:
### What does this PR do?
Add settings.json file (`/remote/data/Machine/settings.json`) to the injector, to be copied in to the checode volume on start.

This settings.json file can be used to configure options for Code, taking precendence over user/in-browser settings, and can be used to set required configuration for running the code editor in Che.

Currently, it is only used to set `"git.defaultCloneDirectory": "/projects/"`

Closes https://github.com/eclipse/che/issues/22653

### How to test this PR?
I've pushed changes to `quay.io/amisevsk/che-code:dev`. Updating a workspace's DevWorkspaceTemplate to reference this image instead of the default injector should result in file `/checode/remote/data/Machine/settings.json` existing on start.

If you use Code to clone a project, it should default the path to `/projects`

### DIFF
--- a/build/remote-config/settings.json
+++ b/build/remote-config/settings.json
@@ -1,0 +1,5 @@
+{
+  // Note: settings in this file will override user-level settings in the workspace.
+  //       Any setting that may be changed by the user should not be set here.
+  "git.defaultCloneDirectory": "/projects/"
+}
--- a/build/dockerfiles/assembly.Dockerfile
+++ b/build/dockerfiles/assembly.Dockerfile
@@ -34,6 +34,7 @@ RUN for f in "/mnt/rootfs/bin/" "/mnt/rootfs/home/che" "/mnt/rootfs/etc/passwd" 
 
 COPY --from=machine-exec --chown=0:0 /go/bin/che-machine-exec /mnt/rootfs/bin/machine-exec
 COPY --chmod=755 /build/scripts/*.sh /mnt/rootfs/
+COPY --chmod=755 /build/remote-config /mnt/rootfs/remote/data/Machine/
 
 # Create all-in-one image
 FROM scratch
--- a/build/scripts/entrypoint-init-container.sh
+++ b/build/scripts/entrypoint-init-container.sh
@@ -16,8 +16,12 @@ cp -r /checode-* /checode/
 # Copy machine-exec as well
 mkdir -p /checode/bin
 cp /bin/machine-exec /checode/bin/
-# copy entrypoint
+# Copy entrypoint
 cp /entrypoint-volume.sh /checode/
+# Copy remote configuration
+mkdir -p /checode/remote
+cp -r /remote /checode
+
 echo "listing all files copied"
 ls -la /checode
 


RomanNikitenko:
it looks like the jobs fail on the `Login to Quay.io` step because of https://github.com/eclipse/che/issues/22660
as far as I know it's actual only for a PR created from a fork

azatsarynnyy (commented):
While testing the proposed changes, there's one thing I noticed that we should be careful with.
Everything works as expected for a newly created workspace. But, when I restart an already created workspace, the existing `/remote/data/Machine/settings.json` file is overwritten by a default one on each workspace start. That means everything the user stored there will be lost after restart.
On the other hand, this is unlikely to cause any real problems as I don't see a case when the user might need to store anything in `/remote/data/Machine/settings.json`. This must be a rather unusual use case 🙂

azatsarynnyy (commented):
Maybe, it would be better to copy the settings file to a new workspace only?
In that case, we will have a nice way to override the default VS Code settings by the Che-specific values keeping the user-modified version untouched.

build/remote-config/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+  // Note: settings in this file will override user-level settings in the workspace.
+  //       Any setting that may be changed by the user should not be set here.
+  "git.defaultCloneDirectory": "/projects/"
```
RomanNikitenko, line 4: @azatsarynnyy @amisevsk 
an alternative way:
put `projects` [here](https://github.com/che-incubator/che-code/blob/6684f6d4ff883edbc05bf6f9950dafa293649cb9/code/extensions/git/package.json#L2348)

this way it's possible for a user to replace the value with another one in settings and it will not be overwritten after restarting a workspace

I don't insist, just FYI

azatsarynnyy: Thanks @RomanNikitenko for mentioning another approach.
Yes, this is how we usually do it, e.g.: https://github.com/che-incubator/che-code/pull/245/files
IMO, a couple of the pros of the approach @amisevsk provided in this PR, is that it allows keeping the VS Code sources untouched and storing all overridden default settings in a single file.

RomanNikitenko (approved):

amisevsk:
Hi all, sorry for the delay in responding. 

Regarding copying once or overwriting on every start (@azatsarynnyy's point), I was unsure the best way to go here. On one hand, if we always copy (and so overwrite the file), any future config settings we add to this file will be applied to existing workspaces -- but we overwrite any user-level changes. 

If we only copy it if it does not exist (i.e. copy only on first start), then user changes persist but if we ever ship a change to this config file it will only apply to new workspaces by default.

Since I figured it's more likely that we change this file on our end, I opted for the first option, but I'm happy to go with either.

build/remote-config/settings.json
amisevsk: I was generally trying to avoid changing anything in a subdir we don't own (I don't want to break the rebase job or anything like that) but it seems like this would be a relatively safe change.

I'm happy to do either here, whatever makes more sense :+1:

amisevsk:
For the PR check, would it make sense to close this PR and open it from a branch in the main repo? I think I have permissions for this.

azatsarynnyy:
> For the PR check, would it make sense to close this PR and open it from a branch in the main repo? I think I have permissions for this.

Unfortunately, until https://github.com/eclipse/che/issues/22660 is fixed, the required PR status check won't work.

I see there are two options currently:
- either open a PR from the branch in this repo
- or wait for @vitaliy-guliy to fix https://github.com/eclipse/che/issues/22660 (AFAIK, he should work on it next week after his returning from his time off)

azatsarynnyy:
> Hi all, sorry for the delay in responding.
> 
> Regarding copying once or overwriting on every start (@azatsarynnyy's point), I was unsure the best way to go here. On one hand, if we always copy (and so overwrite the file), any future config settings we add to this file will be applied to existing workspaces -- but we overwrite any user-level changes.
> 
> If we only copy it if it does not exist (i.e. copy only on first start), then user changes persist but if we ever ship a change to this config file it will only apply to new workspaces by default.
> 
> Since I figured it's more likely that we change this file on our end, I opted for the first option, but I'm happy to go with either.

I think it's a good reason for overwriting the settings on each start of the existing workspace. So, I'm ➕ 1️⃣ for the current implementation 🙂

azatsarynnyy (approved):

amisevsk:
## Moved to PR https://github.com/che-incubator/che-code/pull/314

To make life a little simpler (I hope), I've opened https://github.com/che-incubator/che-code/pull/314 so that we can get PR checks :+1:

PR #311, opened 2023-11-23
Rebase against upstream

RomanNikitenko:
### What does this PR do?
Rebase against the upstream
vscode-upstream-sha1: 27a0cbb26647670ad719bcb47e2d1ca4cee133bf

https://github.com/eclipse/che/issues/22681

### How to test this PR?
- the jobs should be green
- testing of any use cases is useful as we pick up a lot of commits from the upstream


azatsarynnyy (approved):
Thanks @RomanNikitenko for looking at the rebase job and fixing the incompatibilities with the upstream 👍

PR #312, opened 2023-11-23
Cleans up the root `package.json`

### LINKED ISSUE #22686: Che-in-Che development is broken
### Describe the bug

It's not possible to build Che-Code using Che-in-Che development.

### Che version

7.77@latest

### Steps to reproduce

1. Start a workspace using Che-Code repository
2. Run `devfile` commands for the che-in-che development

image

### Expected behavior

it's possible to build Che-Code

### Runtime

I've faced the problem on the dogfooding instance

azatsarynnyy:
### What does this PR do?
In this PR, I suggest removing the redundant fields from the root `package.json` that I overlooked while reviewing [the recently merged patch](https://github.com/che-incubator/che-code/pull/305/files). See the details in the inlined code comments below ⬇️ 

fixes https://github.com/eclipse/che/issues/22686

### How to test this PR?
Actually, the PR changes nothing except cleaning up the part of the root `package.json` that may confuse the developer.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,13 +10,5 @@
     "build:min": "cd code && yarn gulp vscode-reh-web-linux-x64-min",
     "rebuild-native-modules": "cd code && npm rebuild"
   },
-  "license": "EPL-2.0",
-  "dependencies": {
-    "node-gyp": "^9.4.1"
-  },
-  "engines": {
-    "node": ">= 18.18.2",
-    "npm": ">= 9.8.1",
-    "yarn": ">= 1.22.19"
-  }
+  "license": "EPL-2.0"
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,20 +2,3 @@
 # yarn lockfile v1
 
 
-node-gyp@9.4.1:
-  version "9.4.1"
-  resolved "https://registry.npmjs.org/node-gyp/-/node-gyp-9.4.1.tgz#8a1023e0d6766ecb52764cc3a734b36ff275e185"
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -10,13 +10,5 @@
     "build:min": "cd code && yarn gulp vscode-reh-web-linux-x64-min",
     "rebuild-native-modules": "cd code && npm rebuild"
   },
-  "license": "EPL-2.0",
-  "dependencies": {
-    "node-gyp": "^9.4.1"
```
azatsarynnyy, line 15: Removing `node-gyp` from npm was a breaking change introduced in https://github.com/npm/cli/pull/6932.
They already fixed the problem in https://github.com/npm/cli/pull/6932.

While a new version of Node.js with the fixed npm has not yet been released, the only temporary workaround that we need on the Che-Code side is [installing `node-gyp` globally](https://github.com/che-incubator/che-code/pull/305/files#diff-5a132d90967fc45fd3240e3fe2628e15e5ab93f515e9008945e9ccc08f7ab231R51) to allow any transitive dependency rely on it. Once we update to a newer Node.js, we can get rid of that ^^ workaround.

Anyway, if it was even decided that `node-gyp` was no longer part of npm, we don't need to declare any transitive dependency of VS Code in Che-Code's root package.json.

package.json
```diff
@@ -10,13 +10,5 @@
...
-  },
-  "engines": {
```
azatsarynnyy, line 17: VS Code build has a custom check of the node/yarn versions on [the preinstall phase](https://github.com/che-incubator/che-code/blob/main/code/build/npm/preinstall.js). I believe there's not much sense in introducing one more very similar verification. So, I suggest relying only on the verification script from upstream.

RomanNikitenko: moreover: it broke the che-in-che development: https://github.com/eclipse/che/issues/22686

RomanNikitenko (approved):

PR #315, opened 2023-12-01
Check VS Code tests for the ubi9 based image

RomanNikitenko:


RomanNikitenko:
the tests were applied in another PR https://github.com/che-incubator/che-code/pull/324

PR #326, opened 2024-02-01
feat: add devfile walkthrough extension to built-ins

vitaliy-guliy:
### What does this PR do?
Adds vscode devfile extension to built-ins.

Screenshot from 2024-02-01 12-46-15

Screenshot from 2024-02-01 12-46-34

Screenshot from 2024-02-01 12-47-06

! Please note my comment in PR https://github.com/devfile/vscode-walkthrough-extension/pull/3#issuecomment-1919056818

There is a regression in VS Code.
I will provide a small test extension and file the issue to the VS Code.

https://github.com/eclipse/che/issues/21606

### How to test this PR?
- create a workspace with https://github.com/vitaliy-guliy/recommended-extensions-sample/tree/test-pr-326
- open Welcome, go to `Get Started with the Devfile`

### DIFF
--- a/code/product.json
+++ b/code/product.json
@@ -83,7 +83,7 @@
 		},
 		{
 			"name": "devfile.vscode-devfile",
-			"version": "latest",
+			"version": "0.0.2",
 			"repo": "https://github.com/devfile/vscode-walkthrough-extension"
 		}
 	],
--- a/code/product.json
+++ b/code/product.json
@@ -83,7 +83,7 @@
 		},
 		{
 			"name": "devfile.vscode-devfile",
-			"version": "0.0.2",
+			"version": "latest",
 			"repo": "https://github.com/devfile/vscode-walkthrough-extension"
 		}
 	],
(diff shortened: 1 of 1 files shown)


code/product.json
```diff
@@ -80,6 +80,11 @@
 				},
 				"publisherDisplayName": "Microsoft"
 			}
+		},
+		{
+			"name": "devfile.vscode-devfile",
+			"version": "0.0.2",
```
ibuziuk, line 86: would be nice to understand why `latest` does not work

ibuziuk (approved):

code/product.json
vitaliy-guliy: yes, using `latest` here would be the best solution

added comment here https://github.com/eclipse/che/issues/21606#issuecomment-1921457345

l0rd:
Can we make the Devfile walkthrough appear in the welcome page? Ideally as the first Walkthrough.

vitaliy-guliy:
> Can we make the Devfile walkthrough appear in the welcome page? Ideally as the first Walkthrough.

I wanted, but have not found a way how to do it. VS Code plugin API does not provide the ability to change item order. It seems VS Code displays the walkthrouhs due to order in which the extensions are loaded/added to the vscode.

PR #328, opened 2024-02-15
fix: terminal should use '/projects' as default directory

vitaliy-guliy:
### What does this PR do?
If a project or a folder is not opened, uses `/projects` directory as default when opening a terminal.

Solves https://github.com/eclipse/che/issues/22819

### How to test this PR?
[Contribute&message=Dev%20cluster%20(for%20maintainers)&logo=eclipseche&color=525C86&labelColor=FDB940)](https://che-dogfooding.apps.che-dev.x6e0.p1.openshiftapps.com#https://gist.githubusercontent.com/vitaliy-guliy/ba70794781df730cfd603e8d531cbc42/raw/34594b409d439edd07627b23c570dd9d64bade89/devfile.yaml?che-editor=https://gist.githubusercontent.com/vitaliy-guliy/ba952d228e344fff76c0b84c8f85cb90/raw/1c679ec8f53a8ef7a7d573c2ea2a2f4e8c4777c7/devfile.yaml)


azatsarynnyy:
As an alternative, we can consider just adding `"terminal.integrated.cwd": "/projects/"` to https://github.com/che-incubator/che-code/blob/main/build/remote-config/settings.json
In such a case, it will not require modifying Code-OSS sources.
WDYT @vitaliy-guliy @RomanNikitenko

vitaliy-guliy:
> As an alternative, we can consider just adding `"terminal.integrated.cwd": "/projects/"` to https://github.com/che-incubator/che-code/blob/main/build/remote-config/settings.json In such a case, it will not require modifying Code-OSS sources. WDYT @vitaliy-guliy @RomanNikitenko

Would be nice if you write this few days ago, I'm testing it.

RomanNikitenko:
> As an alternative, we can consider just adding `"terminal.integrated.cwd": "/projects/"` to https://github.com/che-incubator/che-code/blob/main/build/remote-config/settings.json In such a case, it will not require modifying Code-OSS sources. WDYT @vitaliy-guliy @RomanNikitenko

in this case `/projects` will be a directory for a terminal for a workspace that is not empty and contains a project
is it expected?

I mean, for example:
- I have `web-app` project
- I expect that `cwd` is `/projects/web-app` in a new terminal
- it looks like `cwd` will be `/projects` if we apply that solution

vitaliy-guliy:
@azatsarynnyy @RomanNikitenko there is a pull request to test the case with applying that "terminal.integrated.cwd" property

https://github.com/che-incubator/che-code/pull/330

RomanNikitenko:
@azatsarynnyy 
could you provide a feedback for https://github.com/che-incubator/che-code/pull/328#issuecomment-1948475177, so we could move on

code/src/vs/workbench/contrib/terminal/browser/terminalInstance.ts
```diff
@@ -410,6 +412,16 @@ export class TerminalInstance extends Disposable implements ITerminalInstance {
...
+		if (!this._workspaceFolder && !this.shellLaunchConfig.cwd) {
+			const envs = this._environmentVariableService.collections.get('eclipse-che.api');
+			if (envs) {
+				const projectsRoot = envs.map.get('PROJECTS_ROOT');
+				if (projectsRoot) {
+					this.shellLaunchConfig.cwd = projectsRoot.value
+				}
+			}
+		}
+
```
RomanNikitenko, line 424: For changes on the VS Code side it would be nice to have the corresponding entities in the `.rebase` folder:
- this prevents your changes from being overwritten by the upstream changes
- it's really helpful when we have to resolve conflicts manually as usually we get a lot of commits from the upstream  and it's hard to track what changes are `che-code` specific one.

If you are not going to provide the corresponding changes for the automatic rebasing - I propose to create a separate file where we could registry `che-code` specific changes, like: PR title and reference for the corresponding pull request. 
wdyt?

vitaliy-guliy: I see there are instructions for replacement, I did not find anything for insertions.

RomanNikitenko: usually we use some VS Code specific line to replace it + add `che-code` specific changes
but I agree it's not obvious and sometimes not easy to adapt target changes to get `rebase`-related entities...

azatsarynnyy:
> @azatsarynnyy could you provide a feedback for [#328 (comment)](https://github.com/che-incubator/che-code/pull/328#issuecomment-1948475177), so we could move on

I agree with Vitaliy's comment https://github.com/che-incubator/che-code/pull/330#discussion_r1492839628

If Code could expand the environment variables set in the settings.json, we could use something like `"terminal.integrated.cwd": "$PROJECT_SOURCE"`. But it doesn't, unfortunately.

RomanNikitenko:
> > @azatsarynnyy could you provide a feedback for [#328 (comment)](https://github.com/che-incubator/che-code/pull/328#issuecomment-1948475177), so we could move on
> 
> I agree with Vitaliy's comment [#330 (comment)](https://github.com/che-incubator/che-code/pull/330#discussion_r1492839628)
> 
> If Code could expand the environment variables set in the settings.json, we could use something like `"terminal.integrated.cwd": "$PROJECT_SOURCE"`. But it doesn't, unfortunately.

in other words my assumption in https://github.com/che-incubator/che-code/pull/328#issuecomment-1948475177 is correct and we can not apply `remote-config/settings.json`-related solution as it doesn't work for another use case that I described above.

OK, so let's close https://github.com/che-incubator/che-code/pull/330 and continue with the current PR to get it merged

RomanNikitenko (approved):

AObuchow:
> > @azatsarynnyy could you provide a feedback for [#328 (comment)](https://github.com/che-incubator/che-code/pull/328#issuecomment-1948475177), so we could move on
> 
> I agree with Vitaliy's comment [#330 (comment)](https://github.com/che-incubator/che-code/pull/330#discussion_r1492839628)
> 
> If Code could expand the environment variables set in the settings.json, we could use something like `"terminal.integrated.cwd": "$PROJECT_SOURCE"`. But it doesn't, unfortunately.

VSCode is actually able to expand environment variables, the syntax is the following: `/${env:$PROJECT_SOURCE}`.

However, this is not a perfect solution: if you start an empty workspace, `$PROJECT_SOURCE` is set to `/projects`. If you then clone a git repo with the VSCode UI, `$PROJECT_SOURCE` is still set to `/projects` (as DWO is responsible for setting this environment variable), and thus, terminals still open to `/projects` instead of `/projects/<cloned-project-name>/`.

I haven't verified whether this PR's solution also has suffers from the same edge-case.

PR #339, opened 2024-03-20
fix: allow the extensions to authenticate

vitaliy-guliy:
### What does this PR do?

It is possible to specify trusted extensions, which can authenticate using your GitHub account by specifying the extensions identifiers in `VSCODE_TRUSTED_EXTENSIONS` environment variable.

```properties
      env:
        - name: VSCODE_TRUSTED_EXTENSIONS
          value: "github.copilot,github.copilot-chat"
```

VS Code launcher will take the content of the environment variable and will create corresponding `trustedExtensionAuthAccess` section in `product.json`

```json
	"trustedExtensionAuthAccess": [
		"github.copilot",
		"github.copilot-chat"
	]
```

Having that, the extensions can easily authenticate without prompting the user.
```
let session = await vscode.authentication.getSession('github', scopes, { silent: true });
```

It is helpful especially if the extension does not provide additional options like this
```
let session = await vscode.authentication.getSession('github', scopes, { createIfNone: true });
```

https://github.com/eclipse/che/issues/22781

Screenshot from 2024-03-27 12-53-15

### How to test this PR?

**!Note:** for the purity of the experiment it is recommended to cleanup the browser database.
(Open browser Developer tools, switch to Application tab and remove everything in `IndexedDB` storage) 

1. Create a workspace with

https://github.com/vitaliy-guliy/recommended-extensions-sample?che-editor=https://gist.githubusercontent.com/vitaliy-guliy/1d4e35e3f6be3b109d4d62948b24b8c6/raw/b975b63aaa5475d3f8442d690353c3c00c9fb864/editor.yaml

or with

https://gist.githubusercontent.com/vitaliy-guliy/baba9c6391f5dc6590b9e7902fdfa6f7/raw/11e46945583df4ccfbb070980ef84c5b993c2671/devfile.yaml?che-editor=https://gist.githubusercontent.com/vitaliy-guliy/1d4e35e3f6be3b109d4d62948b24b8c6/raw/b975b63aaa5475d3f8442d690353c3c00c9fb864/editor.yaml

2. Perform `GitHub: Device Authentication`

3. Download both extensions to your laptop

- [GitHub.copilot-1.173.0.vsix](https://marketplace.visualstudio.com/_apis/public/gallery/publishers/GitHub/vsextensions/copilot/1.173.0/vspackage)

- [GitHub.copilot-chat-0.14.2024032101.vsix](https://marketplace.visualstudio.com/_apis/public/gallery/publishers/GitHub/vsextensions/copilot-chat/0.14.2024032101/vspackage)

4. Open Extensions view, install the extensions

Both extensions should work.

You can check `/checode/checode-linux-libc/ubi8/product.json` file as well to be sure the `trustedExtensionAuthAccess` section is added at the end of the file.

- [x] the PR contains changes in the [code](https://github.com/che-incubator/che-code/tree/main/code) folder (you can skip it if your changes are placed in a che extension )
- [x] the corresponding items were added to the [CHANGELOG.md](https://github.com/che-incubator/che-code/blob/main/.rebase/CHANGELOG.md) file
- [x] rules for automatic `git rebase` were added to the [.rebase](https://github.com/che-incubator/che-code/tree/main/.rebase) folder


vitaliy-guliy:
@musienko-maxim could you please take a look as well?

RomanNikitenko (approved):
Tested: I was able to generate a test and use `explain` function of the extension

image

Great job, @vitaliy-guliy !

launcher/src/trusted-extensions.ts
```diff
@@ -0,0 +1,60 @@
...
+      } else {
+        console.log('  > Unexpected type of trustedExtensionAuthAccess in product.json. Skip this step');
+        return;
+      }
+
+      if (productJSONChanged) {
+        await productJSON.save();
+      }
+    } catch (err) {
+      console.error(`${err.message} Failure to configure truested extensions in produt.json.`);
```
RomanNikitenko: - `truested` => `trusted`
- `produt` => `product`

launcher/src/trusted-extensions.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    if (!env.VSCODE_TRUSTED_EXTENSIONS) {
+      console.log('  > env.VSCODE_TRUSTED_EXTENSIONS is not defined, skip this step');
+      return;
+    }
+
+    try {
+      const extensions = env.VSCODE_TRUSTED_EXTENSIONS.split(',');
+      if (!extensions.length) {
+        console.log('  > env.VSCODE_TRUSTED_EXTENSIONS is empty, skip this step');
```
RomanNikitenko: actually it can be not empty, but another separator is used - so I would rephrase the message + added the value of the env var to the log - it would help to identify the cause of the problem on a user's side just using this log

launcher/src/trusted-extensions.ts
```diff
@@ -0,0 +1,60 @@
...
+        return;
+      }
+
+      for (const e of extensions) {
+        console.log(`  > add ${e}`);
```
RomanNikitenko: add to trusted extensions?
I would rephrase it - what and _where_ it's added

vitaliy-guliy: The whole output in `/checode/entrypoint-logs.txt` should look like this
```
# Configuring Trusted Extensions...
  > add github.copilot
  > add github.copilot-chat
```

launcher/src/trusted-extensions.ts
vitaliy-guliy: I reworked this piece of code

launcher/src/trusted-extensions.ts
```diff
@@ -0,0 +1,68 @@
...
+    if (!env.VSCODE_TRUSTED_EXTENSIONS) {
+      console.log('  > env.VSCODE_TRUSTED_EXTENSIONS is not defined, skip this step');
+      return;
+    }
+
+    try {
+      const extensions: string[] = [];
+
+      const tmp = env.VSCODE_TRUSTED_EXTENSIONS.split(',');
+      for (const e of tmp) {
```
RomanNikitenko: sorry, could you use more readable variables
it's hard to read such code: `e`, `tmp` :-)

launcher/src/trusted-extensions.ts
```diff
@@ -0,0 +1,68 @@
...
+        if (e) {
+          extensions.push(e);
+        }
+      }
+
+      if (!extensions.length) {
+        console.log('  > env.VSCODE_TRUSTED_EXTENSIONS does not specify any extension');
```
RomanNikitenko: I think it makes sense to add the value of the variable to the log
Let's say another separator is applied by a user, then:
- I agree that `extensions` array is empty
- but `env.VSCODE_TRUSTED_EXTENSIONS does not specify any extension` is NOT true
- so it makes sense to clarify it in logs, this way it would help to identify and resolve the problem quickly

launcher/src/trusted-extensions.ts
```diff
@@ -0,0 +1,68 @@
...
+        return;
+      }
+
+      for (const e of extensions) {
+        console.log(`  > add ${e}`);
```
RomanNikitenko: does it makes sense to move the log to the place where an extension name is added to the array?

RomanNikitenko: I mean,
at the moment there is:
- one loop to add an extension
and
- another loop to print a log

launcher/src/trusted-extensions.ts
RomanNikitenko: `tmp` - it's extensions' names?
`e` - extension?

launcher/src/trusted-extensions.ts
vitaliy-guliy: User specified `env.VSCODE_TRUSTED_EXTENSIONS` in his devfile.
Launcher says that `env.VSCODE_TRUSTED_EXTENSIONS` is wrong.
No need here to duplicate the value.

From the other side, if you have an idea, please share it.

launcher/src/trusted-extensions.ts
vitaliy-guliy: done

launcher/src/trusted-extensions.ts
vitaliy-guliy: The code is reworked

launcher/src/trusted-extensions.ts
RomanNikitenko: yes, `Launcher says that env.VSCODE_TRUSTED_EXTENSIONS is wrong.`
but the launcher says nothing about:
- what value of the variable the launcher has got
- what could be wrong
- what is expected

You use here `,` as a separator for the `VSCODE_TRUSTED_EXTENSIONS` env var, 
but, for example, `PATH` env variable uses `:` as a separator

image

If a user uses `:` as a separator (just out of habit) for  the `VSCODE_TRUSTED_EXTENSIONS` then:
-  `VSCODE_TRUSTED_EXTENSIONS` is defined, it's not empty, it contains extensions
- but it doesn't work and there is nothing helpful in logs - isn't it?

My propose was just clarify it in logs - like - if the logic can not handle the variable - then it should be clear:
- env variable `VSCODE_TRUSTED_EXTENSIONS` has `some value here` => 
- using `,` as a separator to parse extensions => 
- can not parse extensions, so the value of `VSCODE_TRUSTED_EXTENSIONS` will be ignored => 
- the expected format of the `VSCODE_TRUSTED_EXTENSIONS` is: `some-value,another-value`

============
anyway I've approved your pull request yesterday, these changes are minor, it's up to you
so feel free to merge the pull request as soon as you believe it's good enough

ibuziuk (approved):
great job
@vitaliy-guliy please, add relevant documentation about the new env var

musienko-maxim (approved):

musienko-maxim:
Was checked on the DevSpaces based on 3.12 @ 0dc45 #10 :: Eclipse Che Dashboard 7.82.0 @ 896c. Worked as expected!

launcher/src/trusted-extensions.ts
vitaliy-guliy: Have just played with it and would say, that the cheapest way in case of errors is to print the current value of `VSCODE_TRUSTED_EXTENSIONS` environment variable and explain that it should contain one or several extension identifiers divided by comma.

It's for the first step.
The next, I think we should document the functionality provided by launcher. Among them, ability to configure trusted extensions, location of Open VSX registry, webviews, etc.

vitaliy-guliy: The output with one failed extension
```
# Configuring Trusted Extensions...
  > env.VSCODE_TRUSTED_EXTENSIONS is set to [redhat.yaml,redhat.openshift,red hat.java]
  > add redhat.yaml
  > add redhat.openshift
  > failure to add [red hat.java] because of wrong identifier
```

The output when the variable does not specify anything
```
# Configuring Trusted Extensions...
  > env.VSCODE_TRUSTED_EXTENSIONS is set to [,,,]
  > ERROR: The variable provided most likely has wrong format. It should specify one or more extensions separated by comma.
```

PR #382, opened 2024-07-08
fix: keep projects when restarting the workspace from local devfile

vitaliy-guliy:

https://github.com/eclipse-che/che/issues/23014

### How to test this PR?
- create workspace from https://github.com/Nykredit/openshift-test/tree/single_root but use the editor image provided by this PR
- once worksace created, open the terminal and check for `PROJECTS_ROOT` and `PROJECT_SOURCE` environment variables
- restart workspace from local devfile
- open the terminal again and check for the variables. Their values should be unshanged
- ensure the color theme is kept

### DIFF
--- a/code/extensions/che-remote/src/extension.ts
+++ b/code/extensions/che-remote/src/extension.ts
@@ -209,18 +209,11 @@ async function updateDevfile(cheApi: any): Promise<boolean> {
   try {
     let projects: V1alpha2DevWorkspaceSpecTemplateProjects[] | undefined = devfileContext.devWorkspace.spec!.template!.projects;
     if (!projects || projects.length === 0) {
-      projects = [];
-  
       const flattenedDevfileContent = await fs.readFile(process.env.DEVWORKSPACE_FLATTENED_DEVFILE!, 'utf8');
       const flattenedDevfile = jsYaml.load(flattenedDevfileContent) as any;
-
       if (flattenedDevfile.projects) {
-        for (const project of flattenedDevfile.projects) {
-          projects.push(project);
-        }
+        devfileContext.devWorkspace.spec!.template!.projects = flattenedDevfile.projects;
       }
-  
-      devfileContext.devWorkspace.spec!.template!.projects = projects;
     }
   } catch (error) {
     await vscode.window.showErrorMessage(`Failed to read Devfile. ${error}`);
--- a/code/extensions/che-remote/src/extension.ts
+++ b/code/extensions/che-remote/src/extension.ts
@@ -187,20 +187,11 @@ async function updateDevfile(cheApi: any): Promise<boolean> {
   const pluginRegistryUrl = process.env.CHE_PLUGIN_REGISTRY_INTERNAL_URL;
   console.info(`Using ${pluginRegistryUrl} to generate a new Devfile Context`);
 
-  // let devfileContent: string;
-  // try {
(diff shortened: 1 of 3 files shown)


code/extensions/che-remote/src/extension.ts
```diff
@@ -202,6 +205,21 @@ async function updateDevfile(cheApi: any): Promise<boolean> {
     return false;
   }
 
+  // if a new Devfile does not contain projects, copy them from flattened Devfile
+  try {
+    let projects: V1alpha2DevWorkspaceSpecTemplateProjects[] | undefined = devfileContext.devWorkspace.spec!.template!.projects;
+    if (!projects || projects.length === 0) {
```
AObuchow, line 211: Something to consider:

When using the [reproducer repo](https://github.com/Nykredit/openshift-test/tree/single_root), if you add  a new project to your devfile, say:
```YAML
projects:
  - name: web-nodejs-sample
    git:
      remotes:
        origin: "https://github.com/che-samples/web-nodejs-sample.git"
```
And then restart from the local devfile, you'll have a single-root workspace that only contains the newly added `web-nodejs-sample` project. The existing project that we used to restart the workspace from local devfile is no longer shown in VSCode, since its project (that used to be present in the devworkspace) wasn't added to the newly generated devworkspace's projects.

So I suggest:
- _Always_ adding the devworkspace's existing projects `(flattenedDevfile.projects)`  to the  newly generated devworkspace's projects (`devfileContext.devWorkspace.spec!.template!.projects`) instead of only adding it _if_ the new devfile does not contain projects.
- Add the devworkspace's existing project(s) to be first in list of projects, since this is the project we had used to restart the workspace from local devfile (it should remain first after we restart)
- Don't add projects with the same name (e.g. if the same project name is used in the devworkspace and local devfile) because this will render the devworkspace invalid.

For reference, the approach I'm suggesting is implemented in DevWorkspace Operator [here](https://github.com/devfile/devworkspace-operator/pull/1193/files#diff-879eb83b19c3051f7f7a0a65d079b4bc52a7891b512757420123a113c03e00c7R91-R123).

vitaliy-guliy: I would give the user all the freedom.
If he added a project to his devfile, he would get only one project in the tree.
Adding a project with the same name should be possible and handled when cloning the project.

In this PR I handled only a case when the user created a workspace from minimal devfile and he wants only to change the devfile a bit and apply the changes.

I would not fix all the cases in this PR. Let's move step by step.
You can check some discussion here https://github.com/eclipse-che/che/issues/22722

RomanNikitenko (approved):
I tested the current PR and can confirm that it fixes the problem that was reported in the https://github.com/eclipse-che/che/issues/23014.

I created a separate issue for the suggestion from @AObuchow (thanks!) - https://github.com/eclipse-che/che/issues/23036.
I reproduced the problem that @AObuchow has described in his comment. I also can reproduce the problem without these changes, so it's no regression. 
So, _I propose_ to provide improvement for the `Restart From Local Devfile` action after discussion in the https://github.com/eclipse-che/che/issues/23036.
But it's up to @AObuchow @vitaliy-guliy  and other reviewers.

code/extensions/che-remote/src/extension.ts
AObuchow: Sorry for slow response but yes this makes sense. If changes are needed, we can discuss them in https://github.com/eclipse-che/che/issues/23036  :+1:

PR #383, opened 2024-07-09
fix: use ~/.bashrc in sh shell

vitaliy-guliy:
### What does this PR do?
- takes all the environment variables from the main vscode thread, saves them into `~/.bashrc_variables` file
- before running any task, applies the variables from `~/.bashrc_variables` file

https://github.com/eclipse-che/che/issues/23009

### How to test this PR?
- Create a workspace using repository https://github.com/vitaliy-guliy/vscode-test-extension/tree/test-sh-terminal and the editor image from this PR [quay.io/che-incubator-pull-requests/che-code:pr-383-amd64](https://quay.io/che-incubator-pull-requests/che-code:pr-383-amd64)
- once workspace created, open a terminal and run `export` command
- launch `Run test script` task. Ensure `PATH` variables are equal for both terminals

### DIFF
--- a/code/extensions/che-terminal/src/extension.ts
+++ b/code/extensions/che-terminal/src/extension.ts
@@ -12,6 +12,8 @@
 
 import * as vscode from 'vscode';
 import { MachineExecClient, TerminalSession } from './machine-exec-client';
+import * as child_process from 'child_process';
+import * as fs from 'fs';
 
 let _channel: vscode.OutputChannel;
 export function getOutputChannel(): vscode.OutputChannel {
@@ -21,9 +23,23 @@ export function getOutputChannel(): vscode.OutputChannel {
 	return _channel;
 }
 
+function saveAllVariables(): void {
+    console.log('>> saving all the environment variables');
+
+    try {
+      const result = child_process.execSync('export');
+      console.log('>> result: ' + result.toString());
+      fs.writeFileSync('/projects/environment-variables.che-terminal', result.toString(), 'utf8');
+    } catch (error) {
+      console.log(`  > Failure to save environment variables. ${error}`);
+    }
+}
+
 export async function activate(context: vscode.ExtensionContext): Promise<Api> {
 	const machineExecClient = new MachineExecClient();
 	await machineExecClient.init();
+
+	saveAllVariables();
 
 	const containers: string[] = await machineExecClient.getContributedContainers();
 
--- a/code/extensions/che-terminal/src/extension.ts
+++ b/code/extensions/che-terminal/src/extension.ts
@@ -30,7 +30,12 @@ function saveAllVariables(): void {
       const result = child_process.execSync('export');
       console.log('>> result: ' + result.toString());
       fs.writeFileSync('/projects/environment-variables.che-terminal', result.toString(), 'utf8');
-    } catch (error) {
+
+      const result2 = child_process.execSync('source ~/.bashrc; export');
+      console.log('>> result2: ' + result2.toString());
+      fs.writeFileSync('/projects/environment-variables.che-terminal-bashrc', result2.toString(), 'utf8');
+
+	} catch (error) {
       console.log(`  > Failure to save environment variables. ${error}`);
     }
 }
@@ -39,7 +44,9 @@ export async function activate(context: vscode.ExtensionContext): Promise<Api> {
 	const machineExecClient = new MachineExecClient();
 	await machineExecClient.init();
 
-	saveAllVariables();
+	setTimeout(() => {
+		saveAllVariables();
+	}, 5000);
 
 	const containers: string[] = await machineExecClient.getContributedContainers();
 
--- a/code/extensions/che-terminal/src/extension.ts
+++ b/code/extensions/che-terminal/src/extension.ts
@@ -40,11 +40,32 @@ function saveAllVariables(): void {
     }
 }
 
+function saveAllVariables1(): void {
+    console.log('>> saving all the environment variables');
+
+    try {
+      const result = child_process.execSync('export');
+      console.log('>> result: ' + result.toString());
+      fs.writeFileSync('/projects/environment-variables.che-terminal.1', result.toString(), 'utf8');
+
+      const result2 = child_process.execSync('source ~/.bashrc; export');
+      console.log('>> result2: ' + result2.toString());
+      fs.writeFileSync('/projects/environment-variables.che-terminal-bashrc.1', result2.toString(), 'utf8');
+
+	} catch (error) {
+      console.log(`  > Failure to save environment variables. ${error}`);
+    }
+}
+
 function saveAllVariables2(): void {
     console.log('>> saving all the environment variables');
 
     try {
-      const result3 = child_process.execSync('source ~/.bashrc; export');
+      const result = child_process.execSync('export');
+      console.log('>> result: ' + result.toString());
+      fs.writeFileSync('/projects/environment-variables.che-terminal.2', result.toString(), 'utf8');
+  
+  	  const result3 = child_process.execSync('source ~/.bashrc; export');
       console.log('>> result3: ' + result3.toString());
       fs.writeFileSync('/projects/environment-variables.che-terminal-bashrc.2', result3.toString(), 'utf8');
 
@@ -57,8 +78,10 @@ export async function activate(context: vscode.ExtensionContext): Promise<Api> {
 	const machineExecClient = new MachineExecClient();
 	await machineExecClient.init();
 
+	saveAllVariables();
+
 	setTimeout(() => {
-		saveAllVariables();
+		saveAllVariables1();
 	}, 5000);
 
 	setTimeout(() => {
--- a/code/extensions/che-terminal/src/extension.ts
+++ b/code/extensions/che-terminal/src/extension.ts
@@ -17,6 +17,8 @@ import * as fs from 'fs';
 import { env } from 'process';
 import * as path from 'path';
 
+export const ENVIRONMENT_VARIABLES = path.join(env.HOME!, '.bashrc_variables');
+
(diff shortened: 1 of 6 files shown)


vitaliy-guliy:
@AObuchow could you take a look as well?

azatsarynnyy (commented):
Thanks @vitaliy-guliy for the suggested patch.

A couple of items I noticed while testing the PR:

1. I got this diff:

image

2. Command `New Terminal (Select a Container)` opens a new terminal that is auto-closed immediately:

https://github.com/user-attachments/assets/d9266b1b-bca3-4123-b85b-7c993cf16ce3

code/extensions/che-terminal/src/machine-exec-client.ts
```diff
@@ -136,6 +136,8 @@ export class MachineExecClient implements vscode.Disposable {
 	 * @returns a TerminalSession object to manage the created terminal session
 	 */
 	async createTerminalSession(component: string, commandLine?: string, workdir?: string, columns: number = 80, rows: number = 24): Promise<TerminalSession> {
+		commandLine = `[ -f ${ENVIRONMENT_VARIABLES} ] && source ${ENVIRONMENT_VARIABLES}; ${commandLine}`;
```
AObuchow: I'm experiencing the same issue @azatsarynnyy described with the "New Terminal (Select Container)" closing immediately. Though it closes quickly, it seems to show something like `sh: command undefined: command not found`.

I believe the problem is coming from the lack of quotes (single or double) surrounding the `commandLine` variable, as sh -c takes a string argument. The use of back ticks ``  `  `` to surround the `commandLine` variable will let TypeScript interpret `commandLine` as a multi-line string, but `commandLine` is not being surrounded by quotes when used in [`cmd`](https://github.com/che-incubator/che-code/blob/eeb3e4aa7a76b848e7cbcb3661770137464d8e8f/code/extensions/che-terminal/src/machine-exec-client.ts#L143). I haven't debugged the code myself however.

If I run the following in a terminal, it will fail
```bash
sh -c [ -f /home/user/.bashrc_variables ] && source /home/user/.bashrc_variables; echo "test"
```

If I instead use single quotes or double quotes instead of a backtick, like the following, it will succeed:

```bash
$ sh -c '[ -f /home/user/.bashrc_variables ] && source /home/user/.bashrc_variables; echo "test"'
test
```

However, I don't think using single or double quotes is necessarily the best option since we don't know whether the `commandLine` function parameter (coming from `createTerminalSession`) will contain single or double quotes, which could close the quotes too early and break things.

A workaround I found (that seems a bit hacky) is to put everything into a stdin for example, running the following in /bin/sh/:
```bash
sh -c "$(cat <<'EOF'
echo "test"
EOF
)"
```
Correctly outputs `test` (i.e. we're running `echo "test"`).

So I believe a potential solution is:

```ts
commandLine = `"$(cat <<'EOF'
[ -f ${ENVIRONMENT_VARIABLES} ] && source ${ENVIRONMENT_VARIABLES}; ${commandLine}
EOF
)"`;
```
Here's a [demonstration](https://www.typescriptlang.org/play/?ssl=5&ssc=3&pln=8&pc=5#code/FAAhBsFMBcQUQHIDUCSAlA8ggsogKgPpICCaKxAQgDJwDKIAvCAEQD0AFgPYC2krArgGdIAJ1YA6AEYBDQexEBjAgDdpIgJbTwkyIOagIMEAp7dpAOwAmVdeciMWkBVxDRd0fQZPczVm3YcAA2YAEgAKBWlYAB5ogHI4DAAxOOAAbRAAWgAzEBCAb0RUTBx8IlJyajoAXxAAXRAAMkaQQU5+RXsCovQsXARCEjJKGlpqgG48-O9fa1tIauBEpOAASmZA8a9OczaocXBOAHMI0ws5u1WgA) of how the output looks in a typescript playground.

AObuchow: There might be a more elegant or safer way to accomplish this, however. I'm not sure my solution is bullet-proof.

vitaliy-guliy:
> A couple of items I noticed while testing the PR:
> 
> 1. I got this diff:

Could you explain a bit what did you compare with what?

azatsarynnyy:
> > A couple of items I noticed while testing the PR:
> > 
> > 1. I got this diff:
> 
> Could you explain a bit what did you compare with what?

@vitaliy-guliy 
I'm sorry, I was comparing the `export`'s output 🤦‍♂️ 
The `PATH`'s values are identical 👍

code/extensions/che-terminal/src/machine-exec-client.ts
vitaliy-guliy: With your final proposal I have
Screenshot from 2024-07-24 15-24-58

Below is the command passed to create a terminal
```
{
  "identifier": {
    "machineName": "dev"
  },
  "cmd": [
    "sh",
    "-c",
    "\"$(cat <<'EOF'\n\t\t\t[ -f /home/user/.bashrc_variables ] && source /home/user/.bashrc_variables\n\t\t\tEOF\n\t\t\t)\""
  ],
  "tty": true,
  "cwd": "",
  "cols": 156,
  "rows": 20
}

```

To be honest I do not understand what's going wrong.
I have tried a dozen of commands but nothing helps, terminal is closed automatically.

AObuchow: I believe the `\t`'s are messing things up here. I'm not sure if they're being converted to actual tabs in the terminal command, or if they're being interpreted as literal characters. Regardless of whether they are tabs or not, they break things when I run the following in my terminal:

```bash
$: "$(cat <<'EOF'
            [ -f /home/user/.bashrc_variables ] && source /home/user/.bashrc_variables
            EOF
        )"
> //... unterminated
```

```bash
$: "$(cat <<'EOF'\n\t\t\t[ -f /home/user/.bashrc_variables ] && source /home/user/.bashrc_variables\n\t\t\tEOF\n\t\t\t)"
bash: warning: command substitution: 1 unterminated here-document
```

You could try removing `\t` from `commandLine` as an extra step maybe? i.e.

```ts
commandLine = `"$(cat <<'EOF'
[ -f ${ENVIRONMENT_VARIABLES} ] && source ${ENVIRONMENT_VARIABLES}; ${commandLine}
EOF
)"`.split("\t").join("");
```

 **All that is important is the new lines must be preserved**, so in theory, the following might work and be safest (since there's no possibility of modifying `${commandLine}` accidentally if it contains `\t`):


```ts
commandLine = `"$(cat <<'EOF'\n[ -f ${ENVIRONMENT_VARIABLES} ] && source ${ENVIRONMENT_VARIABLES}; ${commandLine}\nEOF\n)"`
```

Though trying that in a shell doesn't actually work... but it might work when used as `commandLine`?

vitaliy-guliy: I've played a lot with the command line and noticed that the terminal is closed even if I add a simple echo like
```
commandLine = `echo test;${commandLine}`;
```

If I open a terminal with `New Terminal (Select a Container)`, the terminal receives the empty command
```
{
  "identifier": {
    "machineName": "dev"
  },
  "cmd": [],
  "tty": true,
  "cwd": "",
  "cols": 165,
  "rows": 14
}
```

So my proposal for now is to apply the variables only for tasks, that defined in the devfile.

The command to apply variables with suppressing the output has been improved to

```
commandLine = `test -f ${ENVIRONMENT_VARIABLES} >> /dev/null 2>&1 && source ${ENVIRONMENT_VARIABLES};${commandLine}`;
```

AObuchow: > I've played a lot with the command line and noticed that the terminal is closed even if I add a simple echo 

Weird. Not sure what's going on here.

Regardless, sourcing the .bashrc environment variables only for tasks seems good for https://github.com/eclipse-che/che/issues/23009 as the expected behaviour was:  "PATH on task must have the same value as the PATH env defined inside the image component."

The CI did not seem to build a new editor image with the latest updates to your PR, so I could not test them. However, I tested using [quay.io/che-incubator-pull-requests/che-code:pr-383-amd64](https://quay.io/che-incubator-pull-requests/che-code:pr-383-amd64) from last week with this [devfile](https://github.com/AObuchow/devfile-path-test/blob/main/devfile.yaml) and things seem to work as expected:
- The $PATH in the terminal and task are the same now
- There are no bugs when running a task that contains single or double quotes (though, that's expected -- we only needed to verify that case for the solution I had proposed using `cat << 'EOF'...`)

image

AObuchow (approved):
LGTM & works as expected in my testing :) 

I used `quay.io/che-incubator-pull-requests/che-code:pr-383-amd64` as the editor image and this [devfile](https://github.com/AObuchow/devfile-path-test/blob/main/devfile.yaml) which includes a secondary container component.

I checked the value of $PATH when running as a task inside the secondary container component and it was the same value as $PATH when running inside the normal terminal inside the tooling component:

image

Creating a terminal using `New Terminal (Select a container)` does not close :+1:  Though the value of $PATH is different (as expected currently). However, user's can easily do a `source /home/user/.bashrc_variables` as a workaround for the time being.

Great work @vitaliy-guliy :grin:

code/extensions/che-terminal/src/extension.ts
```diff
@@ -21,10 +27,21 @@ export function getOutputChannel(): vscode.OutputChannel {
 	return _channel;
 }
 
+function saveEnvironmentVariables(): void {
```
RomanNikitenko, line 30: it looks like the function has bad formatting

vitaliy-guliy: No problem, formatting updated.

vitaliy-guliy:
Superseded by https://github.com/che-incubator/che-code/pull/395

PR #393, opened 2024-07-29
Fix Ide not showing

batleforc:
### What does this PR do?

It try to fix the case where the Public key injected for the extension contain a newline by only keeping the part of the pubkey that is indeed the key

I ATM can't build the ide in local, the build process stay stuck for like 30 minute 

https://github.com/eclipse-che/che/issues/23055

### How to test this PR?

1. Add priv/pub ssh key with a newline at the end (automatically added when used with fluxCD ?)
2. Start a Workspace, if the IDE show itself it's good to go

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,5 +1,0 @@
-podman build -f build/dockerfiles/linux-musl.Dockerfile -t linux-musl-amd64 .
-podman build -f build/dockerfiles/linux-libc-ubi8.Dockerfile -t linux-libc-ubi8-amd64 .
-podman build -f build/dockerfiles/linux-libc-ubi9.Dockerfile -t linux-libc-ubi9-amd64 .
-podman build -f build/dockerfiles/assembly.Dockerfile -t harbor.weebo.fr/batleforc/che-code:latest .
-podman push harbor.weebo.fr/batleforc/che-code:latest
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,5 @@
+podman build -f build/dockerfiles/linux-musl.Dockerfile -t linux-musl-amd64 .
+podman build -f build/dockerfiles/linux-libc-ubi8.Dockerfile -t linux-libc-ubi8-amd64 .
+podman build -f build/dockerfiles/linux-libc-ubi9.Dockerfile -t linux-libc-ubi9-amd64 .
+podman build -f build/dockerfiles/assembly.Dockerfile -t harbor.weebo.fr/batleforc/che-code:latest .
+podman push harbor.weebo.fr/batleforc/che-code:latest
--- a/launcher/src/local-storage-key-provider.ts
+++ b/launcher/src/local-storage-key-provider.ts
@@ -61,7 +61,7 @@ export class LocalStorageKeyProvider {
 
     let secret = '';
     for (let i = 0; i < 32; i++) {
-      secret += content.charAt(i * 4);
+      secret += content.charAt(i * 2);
     }
 
     return secret;
--- a/launcher/src/local-storage-key-provider.ts
+++ b/launcher/src/local-storage-key-provider.ts
@@ -28,7 +28,6 @@ export class LocalStorageKeyProvider {
       console.log(`  > found key file ${publicKeyFile}`);
 
       const secret = await this.getPartOfPublicKey(publicKeyFile);
-      console.log(`  > secret: ${secret}`);
       await this.update(FILE_WORKBENCH, SERVER_KEY_MASK, secret);
     } catch (err) {
       console.error(err.message);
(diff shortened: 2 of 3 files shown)


build.sh
```diff
@@ -0,0 +1,5 @@
+podman build -f build/dockerfiles/linux-musl.Dockerfile -t linux-musl-amd64 .
+podman build -f build/dockerfiles/linux-libc-ubi8.Dockerfile -t linux-libc-ubi8-amd64 .
+podman build -f build/dockerfiles/linux-libc-ubi9.Dockerfile -t linux-libc-ubi9-amd64 .
+podman build -f build/dockerfiles/assembly.Dockerfile -t harbor.weebo.fr/batleforc/che-code:latest .
+podman push harbor.weebo.fr/batleforc/che-code:latest
```
ibuziuk: harbor.weebo.fr ?
Do we need this file in general or it was added accidentally

RomanNikitenko: I guess this file should be removed before merging

launcher/src/local-storage-key-provider.ts
```diff
@@ -57,7 +58,7 @@ export class LocalStorageKeyProvider {
 
   async getPartOfPublicKey(file: string): Promise<string> {
     let content = await fs.readFile(file);
-    content = content.substring(content.indexOf(' ') + 1);
+    content = content.split(' ')[1];
```
ibuziuk, line 60: @batleforc could you please add a test for that?

launcher/src/local-storage-key-provider.ts
```diff
@@ -28,6 +28,7 @@ export class LocalStorageKeyProvider {
       console.log(`  > found key file ${publicKeyFile}`);
 
       const secret = await this.getPartOfPublicKey(publicKeyFile);
+      console.log(`  > secret: ${secret}`);
```
vitaliy-guliy: I think it is enough to add a test for that and do not show the secret.
From the other side, I do not mind to keep it here, you can just wrap the secret in quotes to understand whether it contains a newline character.

launcher/src/local-storage-key-provider.ts
vitaliy-guliy: @batleforc It would be great to test it with a key like you mentioned in the issue.

build.sh
batleforc: It was a pure mistake, half of the commit hasn't been pushed and I didn't see it :/

launcher/src/local-storage-key-provider.ts
batleforc: I will add the test, but i can't build the image to test it has a drop-in replacement in my env

batleforc: Test has been added

launcher/tests/local-storage-key-provider.spec.ts
```diff
@@ -19,6 +19,10 @@ const NEW_WORKBENCH_FILE = `
 some code, some code, a mask to be replaced 1234567890ABCDEFGHIJKLMNOPQRSTUV, some code
 `;
 
+const NEW_WORKBENCH_FILE_NEW_LINE = `
+some code, some code, a mask to be replaced ACaZNAIlQqHJOA2XM, some code
+`;
+
```
vitaliy-guliy, line 25: It looks much better now!
Since you are trying to improve getting of a secret, could you please update the `getPartOfPublicKey` method and change
```
secret += content.charAt(i * 4);
```
on
```
secret += content.charAt(i * 2);
```
??

This will allow to take a bit longer secret. So instead of `ACaZNAIlQqHJOA2XM` in the test you can use `AACNa1ZINEAAIblpQoqFHrJPOyAt2gX1`.

vitaliy-guliy: And while googling for possible keyfile format I found an example like this
```
ssh-ed25519
AAAAC3NzaC1lZDI1NTE5AAAAIObHl2pRQsoCqvF7HrrUJwPcOByIAOtS2ggoX51xMWAQ eclipse@che.com
```

But I'm not sure it is a right format.

batleforc: For the key format, ed25519 is the one GitHub seems to recommend https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent

batleforc: The test are fixed with the change i * 2

vitaliy-guliy (approved):
I've tested the editor with the changes and can say it works fine.
@batleforc Good job, thanks!

PR #395, opened 2024-08-05
fix: apply .bashrc when running a task

vitaliy-guliy:
### What does this PR do?
This is alternative to https://github.com/che-incubator/che-code/pull/383

https://github.com/eclipse-che/che/issues/23009

### How to test this PR?

- create a workspace using following
   repository https://github.com/vitaliy-guliy/vscode-test-extension/tree/test-sh-terminal
   editor [quay.io/che-incubator-pull-requests/che-code:pr-395-amd64](https://quay.io/che-incubator-pull-requests/che-code:pr-395-amd64)

- launch test tasks to check the variables for `tools` and `mongo` containers

- check PATH environment variable

The variable seems to be different because of different .bashrs files in both containers.

### DIFF
--- a/code/extensions/che-terminal/src/machine-exec-client.ts
+++ b/code/extensions/che-terminal/src/machine-exec-client.ts
@@ -139,8 +139,7 @@ export class MachineExecClient implements vscode.Disposable {
 	 */
 	async createTerminalSession(component: string, commandLine?: string, workdir?: string, columns: number = 80, rows: number = 24): Promise<TerminalSession> {
 		if (commandLine) {
-			const bashrc = join(env.HOME!, '.bashrc');
-			commandLine = `test -f ${bashrc} >> /dev/null 2>&1 && source ${bashrc};${commandLine}`;
+			commandLine = "test -f ${HOME}/.bashrc >> /dev/null 2>&1 && source ${HOME}/.bashrc >> /dev/null 2>&1;" + commandLine;
 		}
 
 		const createTerminalSessionCall = {
--- a/code/extensions/che-terminal/src/machine-exec-client.ts
+++ b/code/extensions/che-terminal/src/machine-exec-client.ts
@@ -16,6 +16,8 @@ import * as vscode from 'vscode';
 import * as WS from 'ws';
 import { WebSocket } from 'ws';
 import { getOutputChannel } from './extension';
+import { env } from 'process';
+import { join } from 'path';
 
 /** Client for the machine-exec server. */
 export class MachineExecClient implements vscode.Disposable {
@@ -136,6 +138,11 @@ export class MachineExecClient implements vscode.Disposable {
 	 * @returns a TerminalSession object to manage the created terminal session
 	 */
 	async createTerminalSession(component: string, commandLine?: string, workdir?: string, columns: number = 80, rows: number = 24): Promise<TerminalSession> {
+		if (commandLine) {
+			const bashrc = join(env.HOME!, '.bashrc');
+			commandLine = `test -f ${bashrc} >> /dev/null 2>&1 && source ${bashrc};${commandLine}`;
+		}
+
 		const createTerminalSessionCall = {
 			identifier: {
 				machineName: component
--- a/code/extensions/che-terminal/src/machine-exec-client.ts
+++ b/code/extensions/che-terminal/src/machine-exec-client.ts
@@ -16,8 +16,6 @@ import * as vscode from 'vscode';
 import * as WS from 'ws';
 import { WebSocket } from 'ws';
 import { getOutputChannel } from './extension';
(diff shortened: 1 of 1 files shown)


RomanNikitenko (approved):

vitaliy-guliy:
@AObuchow wdyt?

azatsarynnyy (approved):
👍

code/extensions/che-terminal/src/machine-exec-client.ts
```diff
@@ -136,6 +138,11 @@ export class MachineExecClient implements vscode.Disposable {
 	 * @returns a TerminalSession object to manage the created terminal session
 	 */
 	async createTerminalSession(component: string, commandLine?: string, workdir?: string, columns: number = 80, rows: number = 24): Promise<TerminalSession> {
+		if (commandLine) {
+			const bashrc = join(env.HOME!, '.bashrc');
+			commandLine = `test -f ${bashrc} >> /dev/null 2>&1 && source ${bashrc};${commandLine}`;
```
AObuchow: It's worth creating a `.bashrc` that will cause an error (e.g. add `cat nonExistentFile.txt`) to see if that breaks running Devfile tasks with this approach. In theory, it shouldn't since we are using `;` instead of `&&`. However, I suspect any errors in calling `source ~/.bashrc` will be logged when running the Devfile task. I'm not sure if that is desired?

A potential advantage of https://github.com/che-incubator/che-code/pull/383 is that we only call `source ~/.bashrc` once, and [catch the error](https://github.com/che-incubator/che-code/pull/383/files#diff-9b705d84a45991e6f1435dfe87ccee1687896d9c179ede2b08c5a600878a421fR31-R36). 

In the current approach, if the call to `source ~/.bashrc` results in any errors, the errors will repeatedly be hit when running a devfile task.

vitaliy-guliy: I like the idea to experiment with `.bashrc`. Let me then update the repository I'm using to test the pull request.

> However, I suspect any errors in calling source ~/.bashrc will be logged when running the Devfile task. I'm not sure if that is desired?

Good catch. I will experiment with this as well.

vitaliy-guliy:
I found another mistaking.
HOME in mongo container is configured to `/var/lib/mongodb`. But the path to `.bashrc` file was taken based on the HOME taken from the tools container.
```
const bashrc = join(env.HOME!, '.bashrc');
```
As result `.bashrc` from mounted `/home/user` directory will be used as a source.
This is not right.
It needs to get HOME value from `mongo` container and try to find `.bashrc` based on the given value.

AObuchow:
> I found another mistaking. HOME in mongo container is configured to `/var/lib/mongodb`. But the path to `.bashrc` file was taken based on the HOME taken from the tools container.
> 
> ```
> const bashrc = join(env.HOME!, '.bashrc');
> ```
> 
> As result `.bashrc` from mounted `/home/user` directory will be used as a source. This is not right. It needs to get HOME value from `mongo` container and try to find `.bashrc` based on the given value.

Isin't that the intended behaviour for resolving https://github.com/eclipse-che/che/issues/23009 however? I thought we wanted the same $PATH in other containers that the tooling has?

vitaliy-guliy:
> Isin't that the intended behaviour for resolving [eclipse-che/che#23009](https://github.com/eclipse-che/che/issues/23009) however? I thought we wanted the same $PATH in other containers that the tooling has?

If the container has its own `.bashrc`, lets use it.
And it does not prevent us from resolving  [eclipse-che/che#23009](https://github.com/eclipse-che/che/issues/23009)

vitaliy-guliy:
There is no `.bashrc` file in the home directory in mongo container.

To test the behavior I created a `.bashrc` file in project directory and added a test variable.
Then I opened a terminal in mongo container and copied the `.bashrc` to HOME directory with
```
$ cp /projects/vscode-test-extension/.bashrc ${HOME}/
```
and then checked for the file
```
$ ls -la ${HOME}
total 12
drwxrwxr-x. 1 mongodb    root         50 Aug  7 13:47 .
drwxr-xr-x. 1 root       root         21 Sep 17  2020 ..
-rw-r--r--. 1 1000300000 root         31 Aug  7 13:47 .bashrc
-rw-------. 1 1000300000 root        111 Aug  7 13:33 .dbshell
drwxr-xr-x. 2 1000300000 root         20 Aug  7 13:33 .kube
drwxrwsr-x. 4 root       1000300000 4096 Aug  7 13:39 data
```

After that I launched `Run test script - mongo` task and checked the corresponding `.bashrc` is applied

Screenshot from 2024-08-07 16-48-40

AObuchow (approved):
LGTM :+1:

ibuziuk (approved):

PR #411, opened 2024-09-06
fix: load static webview resources with the editor from the same domain

vitaliy-guliy:
### What does this PR do?

Normally, Visual Studio Code loads the webview static resources from a CDN.

After merging this PR the webview static resources in Che-Code ( there are several files https://github.com/che-incubator/che-code/tree/main/code/src/vs/workbench/contrib/webview/browser/pre  ) will be loaded from the same location as the editor ( from the same host ).

Before, we added an option for turning this feature on for instances that are behind a proxy.

Probably something related to the cross-domain policy has been changed in the upstream (but I cannot say for sure) and it brought us some isues with handling the webview events. If the webview scripts are loaded from the same host as the editor, the events are not blocked, so loading everything from one place is the cheapest way to fix the problem.

The feature is still configurable. It can be turned off by defining `WEBVIEW_LOCAL_RESOURCES=false` environment variable for tooling container. But this thing could be discussed. To not confuse the user we can give up using the `WEBVIEW_LOCAL_RESOURCES` variable at all.

https://github.com/eclipse-che/che/issues/22986

### How to test this PR?
1. Any workspace
- create a workspace using the editor image from this pull request
- preview any README that has a hyperlink
- click the link
2. Ansible sample
- create a workspace using the ansible sample with applying the editor image from this pull request
- Openshift welcome page should be opened automatically. Scroll down a bit and click on buttons. Clicking must work.
- Go to Ansible view, click `Get started` and in the appeared `Ansible content creator` tab check that clicking on the links works. 

It would be nice to open any walkthrough that has a markdown (display the content on the right) and check that clicking works as well.

Screenshot from 2024-09-09 11-00-45

### DIFF
--- a/launcher/src/webview-resources.ts
+++ b/launcher/src/webview-resources.ts
@@ -28,8 +28,8 @@ export class WebviewResources {
   async configure(): Promise<void> {
     console.log('# Configuring Webview Resources location...');
 
-    if ('true' !== env.WEBVIEW_LOCAL_RESOURCES) {
-      console.log(`  > env.WEBVIEW_LOCAL_RESOURCES is not set to 'true', skip this step`);
+    if (env.WEBVIEW_LOCAL_RESOURCES !== undefined && 'false' === env.WEBVIEW_LOCAL_RESOURCES) {
+      console.log(`  > env.WEBVIEW_LOCAL_RESOURCES is set to 'false', skip this step`);
       return;
     }
 
--- a/launcher/tests/webview-resources.spec.ts
+++ b/launcher/tests/webview-resources.spec.ts
@@ -13,23 +13,25 @@ import * as path from 'path';
 import * as fs from '../src/fs-extra';
 import { WebviewResources } from '../src/webview-resources';
 
+const getCheCodeEndpointMock = jest.fn();
+jest.mock('../src/flattened-devfile', () => ({
+  FlattenedDevfile: function () {
+    return { getCheCodeEndpoint: getCheCodeEndpointMock };
+  },
+}));
+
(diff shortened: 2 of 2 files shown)


azatsarynnyy (approved):
I tested it on the suggested patch Developer Sandbox instance. Everything works as expected.
See the attached screencast I recorded: 

Screen Recording 2024-09-09 at 12 06 47

azatsarynnyy:
@vitaliy-guliy does it make sense to create an issue for Che-Code currently doesn't support loading from CDN?
In order not to lose this information and make it easy to find for anyone who would be interested in supporting CDN. However, I haven't heard that MS CDN support is required by someone among Che users.

vitaliy-guliy:
> @vitaliy-guliy does it make sense to create an issue for Che-Code currently doesn't support loading from CDN? In order not to lose this information and make it easy to find for anyone who would be interested in supporting CDN. However, I haven't heard that MS CDN support is required by someone among Che users.

It's difficult to say whether or not using the CDN gives some benefits.
I think cashing is not the main reason why Visual Source Code developers decided to use CDN. I think it could be related as well to some iframe security things, but I do not understand which ones. 

FYI, to make it working on our side we patched the html file which is used to initialize the webview iframe.

You can just compare the upstream
https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/webview/browser/pre/index.html#L8
with the downstream one where we removed the sha
https://github.com/che-incubator/che-code/blob/main/code/src/vs/workbench/contrib/webview/browser/pre/index.html#L8

vitaliy-guliy:
@azatsarynnyy thanks for the review!

PR #443, opened 2024-10-23
Add warning message when tracker cannot ping machine-exec

### LINKED ISSUE #23134: Show error if activity tracker extension fails to ping machine-exec

In CheCode, the activity tracker extension sends [events](https://github.com/che-incubator/che-code/blob/ed7034ccc9e09d804e3769b0a64c2e48b81f6c9c/code/extensions/che-activity-tracker/src/extension.ts#L16-L24) to che-machine-exec to let it know that the current workspace is being used by the user. This is done to detect whether the workspace is idle or not, as idle workspaces are terminated in 30 minutes by default, which is configured from `spec.devEnvironments.secondsOfInactivityBeforeIdling` from the CheCluster CR.

If the activity tracker extension fails to ping che-machine-exec, a message is printed in the [extension logs](https://github.com/che-incubator/che-code/blob/ed7034ccc9e09d804e3769b0a64c2e48b81f6c9c/code/extensions/che-activity-tracker/src/activity-tracker-service.ts#L79), however this is not easily noticeable from the user's perspective.

 

If the ping to che-machine-exec fails, a one-time [error notification](https://code.visualstudio.com/api/ux-guidelines/notifications) can be created to let the user know that there are problems with activity/idle detection.

For example:
```
Unable to communicate with idling service due to error: ${error.message}

This workspace may automatically terminate in ${idletimeout} minutes. For ephemeral workspaces, current workspace content may be lost.
 
Please contact an administrator.
```

_No response_

### Additional context

_No response_

dkwon17:
### What does this PR do?
This PR displays an error message to the user if the activity tracker extension fails to ping che-machine-exec. The activity tracker extension regularly pings che-machine-exec to prevent dev environment idling, therefore if the ping fails, this PR lets the user know that their workspace might idle.

Fixes https://github.com/eclipse-che/che/issues/23134

### How to test this PR?

1. Create a workspace with the PR image as the editor URL
2. In the editor, run the following to terminate the che-machine-exec process:
```
// find machine-exec pid
ps -aux

kill -9 <pid>
```
3. Make any file edits in the editor.
4. Wait about 1 minute. After one minute, the error notification should appear:
image
5. Click on "View Logs" to view the error message in the extension's output:

image

### DIFF
--- a/code/extensions/che-activity-tracker/src/activity-tracker-service.ts
+++ b/code/extensions/che-activity-tracker/src/activity-tracker-service.ts
@@ -85,7 +85,8 @@ export class ActivityTrackerService {
 				this.channel.appendLine('Activity tracker: Failed to ping che-machine-exec: ' + error.message);
 				if (!this.errorDisplayed) {
 					this.errorDisplayed = true;
-					this.showErrorMessage();
+					await this.showErrorMessage();
+					this.errorDisplayed = false;
 				}
 			}
 		}
--- a/code/extensions/che-activity-tracker/src/activity-tracker-service.ts
+++ b/code/extensions/che-activity-tracker/src/activity-tracker-service.ts
@@ -7,6 +7,7 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  ***********************************************************************/
+import * as vscode from "vscode";
 
 export type WorkspaceService = { updateWorkspaceActivity: () => any };
 
@@ -28,13 +29,17 @@ export class ActivityTrackerService {
 	private isTimerRunning: boolean;
 	// Flag which is used to check if new requests were received during timer awaiting.
 	private isNewRequest: boolean;
-
+	// Flag used to keep track whether the ping error warning was already displayed or not.
+	private warningDisplayed: boolean;
 	private workspaceService: WorkspaceService;
+	private channel: vscode.OutputChannel;
 
-	constructor(workspaceService: WorkspaceService) {
+	constructor(workspaceService: WorkspaceService, channel: vscode.OutputChannel) {
 		this.isTimerRunning = false;
 		this.isNewRequest = false;
+		this.warningDisplayed = false;
 		this.workspaceService = workspaceService;
+		this.channel = channel;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


RomanNikitenko (approved):
Tested the functionality - it works well for me!

image

code/extensions/che-activity-tracker/src/activity-tracker-service.ts
```diff
@@ -67,17 +72,82 @@ export class ActivityTrackerService {
...
 			if (attemptsLeft > 0) {
-			  setTimeout(this.sendRequest, ActivityTrackerService.RETRY_REQUEST_PERIOD_MS, --attemptsLeft);
+				await new Promise((resolve) => setTimeout(resolve, ActivityTrackerService.RETRY_REQUEST_PERIOD_MS));
+				await this.sendRequest(--attemptsLeft);
 			} else {
-			  console.error('Activity tracker: Failed to ping che-machine-exec: ', error.message);
+				this.channel.appendLine('Activity tracker: Failed to ping che-machine-exec: ' + error.message);
+				if (!this.errorDisplayed) {
+					this.errorDisplayed = true;
+					this.showErrorMessage();
```
vitaliy-guliy: I suppose we have to set `this.errorDisplayed` back to `false` when the popup is closed.
```suggestion
					await this.showErrorMessage();
					this.errorDisplayed = false;
```

vitaliy-guliy (approved):
One small comment about resetting the value of `this.errorDisplayed` variable when popup is closed. Everything else looks good.
Pinging the idling service is restored in a minute and it looks like we need to sow the popup again on failure.

code/extensions/che-activity-tracker/src/activity-tracker-service.ts
dkwon17: @vitaliy-guliy thank you for the review, I didn't set it back to false so that the error message appears only once instead of every minute, WDYT?

RomanNikitenko:
@dkwon17 
PR's checks fail because of the known issue, I've fixed it in the `main` branch,  
so, please fetch recent changes from the `main` branch.

dkwon17:
I've rebased against main and force pushed to my branch

PR #472, opened 2024-12-20
fix: rebuild @devfile/api dependency to fix che-in-che development

vitaliy-guliy:
### What does this PR do?

Re-compiles `@devfile/api` node package on its place inside `node_modules/@devfile/api` directory to have a library with type `commonjs` instead of `module`.

Since the type of `@devfile/api` node package has been changed from `commonjs` to `module`, it become not possible anymore to use `import` from VS Code when we compile it inside the workspace and then launch a secondary editor instance. 

The `@devfile/api` version we are currently using can be found here https://www.npmjs.com/package/@devfile/api/v/2.3.0-1723034342?activeTab=code
It it possible to open `packge.json` and check for `type` property.

The version we used before is https://www.npmjs.com/package/@devfile/api/v/2.2.1-alpha-1667236163?activeTab=code
There is no `type` property in the `package.json` file, so default `commonjs` value is used.

Why do we have the issue when compile and run the editor in the development mode only?

Normally, the `webpack` is used to create the editor bundle, and instead of having a dozen of separate javascript file, we have only one weight, obfuscated javascript file which speeds up the process of downloading the script in the browser.

When we compile che-code in the development mode, the compiler creates thousands of seprarate javascript files and then when we run the backend with node, we face the restriction with the incompatibility of types (vscode is compiled as commonjs).

The changes provided by this pull request will be used only when che-code dogfooding.

Screenshot from 2024-12-20 19-55-42

https://github.com/eclipse-che/che/issues/23163

### How to test this PR?
- create a workspace from this the branch https://github.com/che-incubator/che-code/tree/recompile-devfile-api
- run `devfile: Run VS Code server on port 8000` command. When the command is being executed, it is possible to check the output and see which `@devfile/api` packages have been recompiled.

- run `devfile: Compile with npm` command
- run `devfile: Run VS Code server on port 8000` command and open the secondary editor in a separate window
- once the editor is opened, ensure there are no error notifications appeared

### DIFF
--- a/code/extensions/che-resource-monitor/package-lock.json
+++ b/code/extensions/che-resource-monitor/package-lock.json
@@ -9,7 +9,6 @@
       "version": "0.0.1",
       "license": "EPL-2.0",
       "dependencies": {
-        "@devfile/api": "^2.3.0-1723034342",
         "@kubernetes/client-node": "^0.22.0",
         "inversify": "^6.0.2",
         "reflect-metadata": "^0.2.2",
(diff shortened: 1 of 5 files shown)


vitaliy-guliy:
I think it is better to port changes directly to `@devfile/api`.
A proposed pull request https://github.com/devfile/api/pull/1666

It is possible manually download the script to the project and launch it after installing node dependencies
https://github.com/che-incubator/che-code/blob/recompile-devfile-api/recompile-%40devfile-api.sh

vitaliy-guliy:
Superseded by https://github.com/che-incubator/che-code/pull/506

PR #560, opened 2025-08-27
Support VS Code (local) to Devworkspace over SSH

### LINKED ISSUE #22756: Make it possible to connect to a container via SSH

It is a part of [VS Code desktop flow](https://github.com/eclipse/che/issues/22706), but it is better to decouple as a separate feature.

In general, as a user I want to open a terminal on my laptop and connect to a container in a running workspace via SSH. 

For that we need
- add ssh service ( openssh-server ) to universal developer image
- configure sshd to be able to run in rootless mode
- expose TCP port 21 (or 2121) to allow external connections

rgrunber:
- Fixes https://github.com/eclipse-che/che/issues/22756, based on @vitaliy-guliy 's work at https://github.com/che-incubator/che-code/pull/327 . Unfortunately I don't have permissions to push onto the branch so I've made this separately.

- Custom image based on UDI that contains the non-root SSH daemon, basic
  web server to show webpage on startup to guide the user
- Add custom image to image-publish workflow
- che-code-sshd.yaml file as entrypoint to configure the custom image

To test, all you need is to provide `che-code-sshd.yaml` as the editor definition (you could replace `che-incubator/che-code-sshd` with `rgrunber/che-code-sshd` since the other location doesn't yet exist, and select any workspace sample.

Tried the following configurations :

- :heavy_check_mark:   VS Code with ([`ms-vscode-remote.remote-ssh`](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-ssh)) extension
- :heavy_check_mark: Cursor ("Remote - SSH" embedded) (`anysphere.remote-ssh`) extension
- :heavy_check_mark: VSCodium with ([`jeanp413.open-remote-ssh`](https://open-vsx.org/extension/jeanp413/open-remote-ssh))
- :x: VS Code with ([`jeanp413.open-remote-ssh`](https://open-vsx.org/extension/jeanp413/open-remote-ssh))
  - Initial attempts didn't work, but there might be a way to get this working.
- :heavy_check_mark: Kiro with ([`jeanp413.open-remote-ssh`](https://open-vsx.org/extension/jeanp413/open-remote-ssh))
- :heavy_check_mark:  Artem mentions even a local Jetbrains instance can connect

https://github.com/user-attachments/assets/407d13f4-b10b-475e-b61a-7225c1ab0755


che-code-sshd.yaml
```diff
@@ -0,0 +1,59 @@
+#
```
rgrunber: I assume this file should actually be from this PR and proposed into https://github.com/eclipse-che/che-operator/tree/main/editors-definitions ?

build/scripts/sshd.start
```diff
@@ -0,0 +1,11 @@
...
+
+rm -rf /home/user/ssh
+mkdir -p /home/user/ssh
+cp /opt/ssh/ssh_client_ed25519_key.pub /home/user/ssh/authorized_keys
+
+# start
+/usr/sbin/sshd -D -f /opt/ssh/sshd_config -E /tmp/sshd.log
+
+# stop
+# /usr/sbin/sshd -t -f /opt/ssh/sshd_config -E /tmp/sshd.log
```
rgrunber: Can remove this.

azatsarynnyy:
> Unfortunately I don't have permissions to push onto the branch so I've made this separately.

@rgrunber I've added your GH account to the repository admins list. There should be an invite sent

che-code-sshd.yaml
azatsarynnyy: @rgrunber yes, in that case, it will be available as one more editor option (tile) on the Dashboard

che-code-sshd.yaml
```diff
@@ -0,0 +1,59 @@
...
+    provider: Provided by [Microsoft](https://www.microsoft.com/) under [License](https://code.visualstudio.com/License)
+    title: Visual Studio Code server for Eclipse Che over SSH - latest
+    repository: https://github.com/che-incubator/che-code
+    firstPublicationDate: '2025-08-01'
+
+components:
+  - name: che-code-sshd
+    container:
+      image: quay.io/che-incubator/che-code-sshd:latest
+      memoryLimit: 6144Mi
```
azatsarynnyy: I think, for this container, it should be enough to set the resource requirements the same as for the regular Che-Code container:
https://github.com/eclipse-che/che-operator/blob/980fe0ef9ed90e76e2891deacedb0dab9a8e4e90/editors-definitions/che-code-latest.yaml#L102-L105

azatsarynnyy (approved):
I can confirm the solution works as expected:

Image

che-code-sshd.yaml
rgrunber: I'll go ahead and do this. I've addressed the other issues (and fixed the sshd.pid file, to be written correctly).

rgrunber: Ok, removed it from this PR and re-located to https://github.com/eclipse-che/che-operator/pull/2039 .

build/scripts/server.js
```diff
@@ -0,0 +1,59 @@
...
+Host localhost
+  HostName 127.0.0.1
+  User dev
+  Port 2022
+  IdentityFile /path/to/the/ssh_client_ed25519_key
+        </pre>
+        </p>
+        </li>
+      </ol>
+    </div>
```
rgrunber, line 66: I hope we can reduce the number of steps here down the road (be logged in, oc port-forward, config + identity file). However, there is one problem that can arise that is worth mentioning. The image's host keys are set at image build-time, so whenever the Dockerfile changes (updates, improvements), they would be regenerated (I don't think the build process caches parts of the Dockerfile from previous steps that remain unchanged), and ultimately, users would fail to connect with :

```
    @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
    @    WARNING: REMOTE HOST IDENTIFICATION HAS CHANGED!     @
    @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
```

The solution is just to remove the localhost/127.0.0.1 entries saved in `$HOME/.ssh/known_hosts` but I think it's worth mentioning this as a troubleshooting step here. Maybe even the logfile to check if there are any connection issues.

rgrunber: Added some info about this.

azatsarynnyy: I'm not sure, but maybe it's possible to use a key stored in a workspace:

image

Just an idea to look at.

che-code-sshd.yaml
rgrunber: I think we may have to revisit this. The Java language server along with the extension are all running in the container, not on the user's host system. As a result, the container really needs sufficient resources to run. When I imported a project, it would run fairly slowly until I bumped resources to :

```diff
diff --git a/che-code-sshd.yaml b/che-code-sshd.yaml
index 71ff72c8552..4ea5f2cb308 100644
--- a/che-code-sshd.yaml
+++ b/che-code-sshd.yaml
@@ -36,10 +36,10 @@ components:
   - name: che-code-sshd
     container:
       image: quay.io/rgrunber/che-code-sshd:latest
-      memoryLimit: 1024Mi
+      memoryLimit: 4096Mi
       memoryRequest: 256Mi
-      cpuLimit: 500m
-      cpuRequest: 30m
+      cpuLimit: 1000m
+      cpuRequest: 500m
```

rgrunber:
@azatsarynnyy , I don't have anything else to do here for now, so feel free to merge at earliest convenience.

PR #578, opened 2025-10-01
CRW-9383: only root perms for /etc/passwd

sbouchet:
### What does this PR do?
similar to https://github.com/che-incubator/jetbrains-ide-dev-server/pull/53
[entrypoint.sh](https://github.com/che-incubator/che-code/blob/main/build/scripts/entrypoint.sh) should still add the user to `/etc/passwd`

https://issues.redhat.com/browse/CRW-9383

### How to test this PR?
using `go-runtime` sample, using the built images, open terminal in che-code, then verify that `/etc/passwd` as correct permissions : 
```
bash-5.1$ stat -c '%A %U %G' /etc/passwd
-rw-r--r-- root root
```
and user is correctly inserted in the file
```
bash-5.1$ whoami
1001660000
bash-5.1$ cat "/etc/passwd"
root:x:0:0:root:/root:/bin/bash
...
1001660000:x:1001660000:0:1001660000 user:/opt/app-root/src:/sbin/nologin
```
finally, even if the user shell is set to 'nologin', verify that the 'che-code' env variable is set to proper value:
```
bash-5.1$ echo $SHELL
/bin/bash
```


RomanNikitenko (commented):
@sbouchet 
I guess similar changes are required for the downstream

sbouchet:
With this changes ( revert permissions, removed boilterplate ) and the `go-runtime` sample, che-code can start, open a terminal and etc/paswwd permissions are ok.
SHELL env variable is also set to `/bin/bash` to avoid issues.
other examples using a different container image with other permissions still works as this fix as no impact on them.

RomanNikitenko (approved):

PR #651, opened 2026-02-18
ci: Fix publishing images

RomanNikitenko:
Generated-by: Cursor AI

### What does this PR do?

This change fixes the publish job in `.github/workflows/image-publish.yml` failing after merges to main.
The workflow was calling `docker manifest annotate` for `amd64-*` and `arm64-*` source tags. Those source tags are now manifest lists, and annotate fails with:
`<image>:amd64-<tag> is a manifest list`.

Since platform metadata is already present from the Buildx `--platform ... --push` flow, explicit annotation is no longer needed. The publish step now relies on `docker manifest create --amend ... --amend ...` followed by `docker manifest push`, which is compatible with current source image format.

https://github.com/eclipse-che/che/issues/23745

### How to test this PR?
- [ ] Merge to main and verify image-publish passes.
- [ ] Confirm publish step no longer fails with is a manifest list.
- [ ] Verify resulting tags are published:
quay.io/che-incubator/che-code:next
quay.io/che-incubator/che-code:insiders
quay.io/che-incubator/che-code:insiders-<short_sha>
- [ ] Inspect manifests and confirm both platforms are present (linux/amd64, linux/arm64).

Check manifest command: `podman manifest inspect quay.io/che-incubator/che-code:next`

### DIFF
--- a/.github/workflows/image-publish.yml
+++ b/.github/workflows/image-publish.yml
@@ -141,28 +141,18 @@ jobs:
           SHORT_SHA1=$(git rev-parse --short=7 HEAD)
           if [[ ${{env.BRANCH_NAME}} == main ]]; then         
             docker manifest create quay.io/che-incubator/che-code:next --amend quay.io/che-incubator/che-code:amd64-${{env.tag_suffix}} --amend quay.io/che-incubator/che-code:arm64-${{env.tag_suffix}}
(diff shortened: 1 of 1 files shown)


RomanNikitenko:
it looks like the problem was fixed on the github side.
I was keeping this PR open for the case if the problem come back, but no problem during couple weeks - so I'm closing the current PR...

PR #653, opened 2026-02-18
Add support for UBI 8 based images in SSHD service

rgrunber:
This PR is meant to work with the editor definition proposed in https://github.com/eclipse-che/che-operator/pull/2088

- Fixes https://issues.redhat.com/browse/CRW-10044
- Also discovered that under certain conditions, the random user is not added under `/etc/passwd`. Before, we would use NSS_WRAPPER to modify the home directory and shell stored for a user (under `/etc/passwd`), but now we've expanded to also adding the user if not present.


build/dockerfiles/assembly.sshd.Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+# UBI 8
+FROM registry.access.redhat.com/ubi8/nodejs-20-minimal:latest as sshd-ubi8
+
+USER 0
+
+RUN microdnf -y install libsecret openssh-server nss_wrapper-libs && \
+    microdnf -y clean all --enablerepo='*'
+
+# UBI 9/10
+FROM registry.access.redhat.com/ubi9/nodejs-20-minimal:9.6
```
rgrunber: Consider bumping to `9.7`.

build/dockerfiles/assembly.sshd.Dockerfile
```diff
@@ -0,0 +1,46 @@
+# Copyright (c) 2025 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
+# UBI 8
+FROM registry.access.redhat.com/ubi8/nodejs-20-minimal:latest as sshd-ubi8
```
rgrunber: We don't need node here, just the sshd tooling, so maybe try `ubi8/ubi-minimal:8.10`

azatsarynnyy (approved):
I tested the PR by providing the following editor definition to the Dashboard
https://gist.githubusercontent.com/azatsarynnyy/56a07b64fbd13fac38844baee005971f/raw/43d3517bfdf0aec892e4c63f72f5abf798090f93/editor-over-ssh

It's based on the https://github.com/eclipse-che/che-operator/pull/2088 with the editor image built from the current PR and pushed to quay.io/azatsary/ssh:pr-653

A new workspace, created from the `Node.js Runtime` sample (which is UBI8-based), started as expected, and local IDE connect:

Image

Roland, thanks for moving the landing page startup command to the `sshd.init` script. It makes the solution clearer and more flexible. So, the command can be modified w/o introducing the changes to the editor definition yaml files.

PR #793, opened 2026-08-19
fix: handle ENOENT when scanning TCP proc files

### LINKED ISSUE #23619: Activating extension 'eclipse-che.port' failed: ENOENT: no such file or directory, open '/proc/net/tcp6'.
### Describe the bug

Endpoints do not load when the operating system has IPv6 disabled. This is the error that appears: "Activating extension 'eclipse-che.port' failed: ENOENT: no such file or directory, open '/proc/net/tcp6'."

I include the two references to the source code where I believe the error is:

https://github.com/che-incubator/che-code/blob/main/code/extensions/che-port/src/port-scanner.ts#L59

https://github.com/che-incubator/che-code/blob/main/code/src/vs/workbench/api/node/extHostTunnelService.ts#L255

### Che version

7.110@latest

### Steps to reproduce

1. Load with ipv6 disable or folder /proc/net/tcp6 not exists

### Expected behavior

The editor should work correctly if the operating system works only in IPv4 or IPv6.

### Runtime

Kubernetes (vanilla)

### Screenshots

Image

Image

### Installation method

OperatorHub

### Environment

Linux

### Eclipse Che Logs

```shell

```

### Additional context

_No response_

msivasubramaniaan:
### What does this PR do?

Handles missing /proc/net/tcp and /proc/net/tcp6 files in the Che port scanner.

When IPv4 or IPv6 is disabled, the corresponding proc file may not exist, causing the port scanner to fail with an ENOENT error. This change treats ENOENT as an unavailable protocol and continues scanning ports from the available protocol.

When both IPv4 and IPv6 are available, both files continue to be scanned as before.

Fixes [Eclipse Che #23619](https://github.com/eclipse-che/che/issues/23619).

### How to test this PR?

- Run the existing port scanner tests.
- Verify that ports are detected when both IPv4 and IPv6 proc files are available.
- Verify that port scanning continues successfully when /proc/net/tcp6 is unavailable.
- Verify that port scanning continues successfully when /proc/net/tcp is unavailable.
- Verify that an empty result is returned when both proc files are unavailable.
- Verify that errors other than ENOENT are still propagated.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved listening-port detection across IPv4 and IPv6 connections.
  * Port scanning now continues when one or both protocol data sources are unavailable.
  * Unexpected scanning errors continue to be reported instead of being silently ignored.

* **Tests**
  * Added coverage for partial, complete, and unexpected port-data availability scenarios.

### DIFF
--- a/code/extensions/che-port/src/port-scanner.ts
+++ b/code/extensions/che-port/src/port-scanner.ts
@@ -54,9 +54,12 @@ export class PortScanner {
    */
   public async getListeningPorts(): Promise<ListeningPort[]> {
     const ipConverter = new IpConverter();
-    const outIPV6 = this.scanner.getListeningPortV6();
-    const outIPV4 = this.scanner.getListeningPortV4();
-    const output = (await Promise.all([outIPV4, outIPV6])).join();
+    const [outputIPV4, outputIPV6] = await Promise.all([
+      this.readListeningPorts(() => this.scanner.getListeningPortV4()),
+      this.readListeningPorts(() => this.scanner.getListeningPortV6()),
+    ]);
+
+    const output = outputIPV4 + outputIPV6;
 
     // parse
     const regex = /:\s(.*):(.*)\s[0-9].*\s0A\s/gm;
@@ -72,5 +75,16 @@ export class PortScanner {
       ports.push(port);
     }
     return ports;
+  }
+
+  private async readListeningPorts(reader: () => Promise<string>): Promise<string> {
+    try {
+      return await reader();
+    } catch (error) {
+      if (error?.code === 'ENOENT') {
+        return '';
+      }
+      throw error;
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


msivasubramaniaan:
@rgrunber as discussed, I don't have Linux machine. please test and review the PR

code/extensions/che-port/src/port-scanner.ts
```diff
@@ -73,4 +76,15 @@ export class PortScanner {
     }
     return ports;
   }
+
+  private async readListeningPorts(reader: () => Promise<string>): Promise<string> {
+    try {
+      return await reader();
+    } catch (error) {
+      if (error?.code === 'ENOENT') {
```
rgrunber, line 84: Is the optional chaining on `error` needed ? If the catch is triggered, it cannot be `undefined`.

rgrunber (approved):
Seems fine, I just haven't been able to get the testsuite to run without issues.

code/extensions/che-port/src/port-scanner.ts
msivasubramaniaan: Yes, I think the optional chaining is useful here. Although the `catch` block is triggered, JavaScript allows anything to be thrown, including `undefined` or `null`. So `error?.code` safely handles those cases without throwing another error.

PR #799, opened 2026-09-08
SSHD startup script should refresh environment variables on restart

batleforc:
### What does this PR do?

Create a new file ($HOME/.profile) in the home folder that will get the env var from the process and  source it from .bashrc. 

https://github.com/eclipse-che/che/issues/23947

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved environment setup during SSH sessions by consistently updating the dedicated SSH profile with current environment variables and PATH settings.
  * Ensured `.profile` and `.bashrc` independently load the SSH profile when needed.
  * Streamlined profile validation and updates to keep environment configuration consistent.

### DIFF
--- a/build/scripts/sshd.start
+++ b/build/scripts/sshd.start
@@ -71,12 +71,20 @@ mkdir -p $HOME/bin
 cp $sshd_libdir/tar $sshd_libdir/gzip $sshd_libdir/which $HOME/bin/
 
 # Set up environment variables injected into PID 1 (.profile & .bashrc)
-grep -q 'export DEVWORKSPACE_NAME=' $HOME/.profile
+
+grep -q 'source $HOME/.sshd_profile' $HOME/.profile
 if [ ! $? -eq 0 ]; then
-  env | sed 's|^|export |' >> $HOME/.profile
-  echo 'export PATH=$PATH:$HOME/bin' >> $HOME/.profile
-  cat $HOME/.profile >> $HOME/.bashrc
+  touch $HOME/.sshd_profile
+  echo "source $HOME/.sshd_profile" >> $HOME/.profile
+  echo "source $HOME/.sshd_profile" >> $HOME/.bashrc
 fi
+
+if [ -f $HOME/.sshd_profile ]; then
+  env | sed 's|^|export |' >> $HOME/.sshd_profile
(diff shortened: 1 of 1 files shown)


build/scripts/sshd.start
```diff
@@ -71,12 +71,20 @@ mkdir -p $HOME/bin
...
 
 # Set up environment variables injected into PID 1 (.profile & .bashrc)
-grep -q 'export DEVWORKSPACE_NAME=' $HOME/.profile
-if [ ! $? -eq 0 ]; then
-  env | sed 's|^|export |' >> $HOME/.profile
-  echo 'export PATH=$PATH:$HOME/bin' >> $HOME/.profile
-  cat $HOME/.profile >> $HOME/.bashrc
-fi
+
+touch $HOME/.sshd_profile
```
rgrunber: I don't think we need `touch` here. Both redirecting standard output whether overwriting or appending will create the file.

build/scripts/sshd.start
```diff
@@ -71,12 +71,20 @@ mkdir -p $HOME/bin
...
+
+# Each file is validated independently, so a missing source line in either one
+# (or a deleted .sshd_profile) is repaired instead of skipping the whole block.
+for rc in $HOME/.profile $HOME/.bashrc; do
+  if ! grep -qsF '.sshd_profile' "$rc"; then
+    echo 'source $HOME/.sshd_profile' >> "$rc"
+  fi
+done
+
+echo "" > $HOME/.sshd_profile
```
rgrunber: How about removing this `echo "" > $HOME/.sshd_profile` call and then below change the first `>>` to `>`. It should have the same effect as wiping the file.

rgrunber (changes requested):
I'm aware that this preserved some stale environment variables so this seems like the right thing to do.

build/scripts/sshd.start
batleforc: It should be good, i merged your two recommendation

build/scripts/sshd.start
rgrunber: @batleforc , I still see `touch` here. It should be possible to just remove it.

rgrunber (approved):
Ok, I've updated the PR. I think this is good to merge. I'm also aware of another bug with the environment variables :

If one contains a space, the sourcing can fails, so I have a separate PR to address that.

PR #792, opened 2026-08-18
fix: invalidate restored device auth sessions after token removal

msivasubramaniaan:
### What does this PR do?

Fixes GitHub Copilot Device Authentication sessions being restored by VS Code after the Device Authentication token has been removed.

The PR tracks authentication sessions created using Device Authentication in VS Code SecretStorage. During workspace restart, persisted Device Authentication sessions are removed when Device Authentication is no longer active.

The provider also avoids immediately recreating a session using a fallback PAT/git-credential token after removing the persisted Device Authentication session, requiring the user to authenticate again.

https://github.com/user-attachments/assets/0bcd87c0-5d14-49b3-becc-eaec42ca3fbf

### What issues does this PR fix?

Fixes the issue where removing the GitHub Copilot Device Authentication token from one workspace does not invalidate the persisted authentication session in other existing workspaces.

After removing the Kubernetes `device-authentication-secret-*` secret, another workspace could restore the previously persisted VS Code authentication session after restart and continue using GitHub Copilot without re-authentication.

### How to test this PR?

- Open User Dashboard in browser in incognito mode as a non-admin user. 
- Create workspace WS1 and complete GitHub Device Authentication flow - verify Copilot Chat works. 
- Create workspace WS2 - verify Copilot Chat works without authentication (token shared).\
- In WS2, open Command Palette (F1) and run GitHub Copilot: Remove Device Authentication Token
- Verify device-authentication-secret-* k8s secret is deleted from user namespace
- Restart WS1 and open Copilot Chat

### Does this PR contain changes that override default upstream Code-OSS behavior?
- [ ] the PR contains changes in the [code](https://github.com/che-incubator/che-code/tree/main/code) folder (you can skip it if your changes are placed in a che extension )
- [ ] the corresponding items were added to the [CHANGELOG.md](https://github.com/che-incubator/che-code/blob/main/.rebase/CHANGELOG.md) file
- [ ] rules for automatic `git rebase` were added to the [.rebase](https://github.com/che-incubator/che-code/tree/main/.rebase) folder

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Device Authentication session tracking and persistence across restarts.
  * Prevented removed, invalid, or incomplete sessions from being restored unexpectedly.
  * Ensured session records are cleared consistently during sign-out, cleanup, invalidation, and authentication changes.
  * Failed session tracking now safely rolls back incomplete authentication sessions.
  * Copilot now clearly explains when GitHub authentication is required.
[...]

PR #808, opened 2026-09-12
fix: handle missing device auth token during session hydration

msivasubramaniaan:
### What does this PR do?
This PR resolves the regression of https://github.com/che-incubator/che-code/pull/792

https://redhat.atlassian.net/browse/CRW-13027

https://github.com/user-attachments/assets/9e382804-6a78-49fd-a350-1a2e5e06d6d4

## Bug Fixes

- Improved GitHub authentication reliability when device-authentication session tracking fails.
- Prevented incomplete sessions from being persisted when device-authentication status cannot be saved or retrieved.
- Automatically rolls back session changes when device-authentication tracking encounters an error.

### DIFF
--- a/code/extensions/che-github-authentication/src/github.ts
+++ b/code/extensions/che-github-authentication/src/github.ts
@@ -353,20 +353,20 @@ export class GitHubAuthProvider implements vscode.AuthenticationProvider {
 
     await this.storeSessions(updatedSessions);
     if (isDeviceAuth) {
-      const deviceAuthSessionIds = await this.getDeviceAuthSessionIds();
+      try {
+        const deviceAuthSessionIds = await this.getDeviceAuthSessionIds();
 
-      if (!deviceAuthSessionIds.includes(session.id)) {
-        try {
+        if (!deviceAuthSessionIds.includes(session.id)) {
           await this.storeDeviceAuthSessionIds([
             ...deviceAuthSessionIds,
             session.id,
           ]);
-        } catch (error) {
-          // Roll back the session because its Device Authentication
-          // tracking ID could not be persisted.
-          await this.storeSessions(sessions);
-          throw error;
         }
+      } catch (error) {
+        // Roll back the session because its Device Authentication
+        // tracking ID could not be persisted.
+        await this.storeSessions(sessions);
+        throw error;
       }
     }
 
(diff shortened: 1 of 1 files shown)


RomanNikitenko (commented):
@msivasubramaniaan 
Could you add information about the fix?
What was the cause of the problem and how current changes are going to fix it? 
What specific error occurred when the extension failed to activate?

I'm asking because I'm afraid adding try/catch can hide a problem - so the extension will be activated, but a new problem is going to be hidden

msivasubramaniaan:
> @msivasubramaniaan Could you add information about the fix? What was the cause of the problem and how current changes are going to fix it? What specific error occurred when the extension failed to activate?
> 
> I'm asking because I'm afraid adding try/catch can hide a problem - so the extension will be activated, but a new problem is going to be hidden

Hello @RomanNikitenko
The regression was caused by the new isDeviceAuthToken() check introduced at https://github.com/che-incubator/che-code/blob/e2e91b7085d621233bedddfcbd00acd303b8ec32/code/extensions/che-github-authentication/src/github.ts#L87. Since this check was not handling the failure case, an error while determining the authentication state propagated through the hydration flow and affected extension activation. The latest [PR](https://github.com/che-incubator/che-code/pull/808) now handles this failure case gracefully, preventing the authentication-state check from impacting extension activation and thereby resolving the regression. I have completely tested the fix and the issue was resolved

RomanNikitenko:
@msivasubramaniaan 
could you share error message 
I mean - what specific error occurred when the extension failed to activate?

msivasubramaniaan:
> @msivasubramaniaan could you share error message I mean - what specific error occurred when the extension failed to activate?

Here is the error:
`Error: Activating extension 'vscode.che-github-authentication' failed: GitHub authentication token is not setup.`

msivasubramaniaan:
Hello @RomanNikitenko,

The main root cause of the regression is the new token access introduced in the hydration flow. The code calls `getToken()` during initialization, but the token may not be available yet at that point.

In the earlier implementation, this case was explicitly handled by wrapping `getToken()` and the hydration logic in a `try/catch`. If the token was not available after initialization, the error was caught and the hydration flow continued safely:
https://github.com/che-incubator/che-code/blob/7553515130a094a13de9b03dd6a97959be37445b/code/extensions/che-github-authentication/src/github.ts#L109-L115

In the new implementation, the token is accessed directly during `hydrateFromK8sToken()` without handling the case where the token is not yet available. Since this method is part of the extension activation flow, the resulting error propagated and caused the activation regression.

The latest PR addresses this by properly handling the failure case, so the extension can continue its activation flow even when the token is not available immediately after initialization.

